PR #7, opened 2013-05-15
Adding roslaunch style if/unless conditional statements to xacro

piyushk:
This patch allows adding conditionals to xacro with the following syntax. This is useful in the case of configurable robots or loading different gazebo plugins.

```
<xacro:if value="<expression/substitution_arg>">
  <... some xml code here ...>
</xacro:if>
<xacro:unless value="<expression/substitution_arg>">
  <... some xml code here ...>
</xacro:unless>
```

The expression needs to evaluate to "0", "true", "1" or "false", otherwise an error will be thrown. Additionally, the conditional has no effect on macro or property definitions, and is only used while constructing the final xml.

The pull request also includes a minor bug fix.

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -504,6 +504,26 @@ def eval_all(root, macros, symbols):
                     raise XacroException("Block \"%s\" was never declared" % name)
 
                 node = None
+            elif node.tagName == 'if' or node.tagName == 'xacro:if':
+                value = eval_text(node.getAttribute('value'), symbols)
+                if value == 1 or value == 'true':
+                    for e in list(child_elements(node)):
+                        cloned = node.cloneNode(deep = True)
+                        eval_all(cloned, macros, symbols)
+                        node.parentNode.insertBefore(e, node)
+                elif value != 0 and value != 'false':
+                    raise XacroException("Xacro conditional evaluated to \"%s\". Acceptable evaluations are one of [\"1\",\"true\",\"0\",\"false\"]" % value)
+                node.parentNode.removeChild(node)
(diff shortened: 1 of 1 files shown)


jonbinney:
This looks very useful and I can't think of anything it will break; anyone have any thoughts on this? /cc @chadrockey @davetcoleman @jon-weisz @isucan

src/xacro/__init__.py
```diff
@@ -504,6 +504,26 @@ def eval_all(root, macros, symbols):
                     raise XacroException("Block \"%s\" was never declared" % name)
 
                 node = None
+            elif node.tagName == 'if' or node.tagName == 'xacro:if':
+                value = eval_text(node.getAttribute('value'), symbols)
```
isucan, line 508: I am not familiar with this code, but can this ever be None? I'm wondering whether an exception could ever be triggered by the next if statement.

isucan:
This looks good indeed! Is there a strong reason to have this in groovy? I really think we should leave this for hydro (and the pull request should be made for hydro-devel instead of groovy-devel). We need to update the docs on the wiki when merging this.

src/xacro/__init__.py
piyushk, line 508: I don't think node can ever be None here. A similar check for insert_block is right above this one.

piyushk:
+1 for checking this into hydro-devel. I can't remember why I submitted the pull request to groovy-devel. Probably didn't think too much about it at the time. Should I resubmit?

isucan:
Yes, please resubmit for the hydro branch

jonbinney:
Closing this; when you pull request to the hydro-devel branch I'll merge that.

PR #8, opened 2013-06-02
xacro conditional blocks

piyushk:
resubmitting pull request https://github.com/ros/xacro/pull/7 to be merged into hydro-devel

---

This patch allows adding conditionals to xacro with the following syntax. This is useful in the case of configurable robots or loading different gazebo plugins.

```
<xacro:if value="<expression/substitution_arg>">
  <... some xml code here ...>
</xacro:if>
<xacro:unless value="<expression/substitution_arg>">
  <... some xml code here ...>
</xacro:unless>
```

The expression needs to evaluate to "0", "true", "1" or "false", otherwise an error will be thrown. Additionally, the conditional has no effect on macro or property definitions, and is only used while constructing the final xml.

The pull request also includes a minor bug fix. https://github.com/piyushk/xacro/commit/5c0cbd7e939bf8dc782d2914003d7ffd3f25788f

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -504,6 +504,26 @@ def eval_all(root, macros, symbols):
                     raise XacroException("Block \"%s\" was never declared" % name)
 
                 node = None
+            elif node.tagName == 'if' or node.tagName == 'xacro:if':
+                value = eval_text(node.getAttribute('value'), symbols)
+                if value == 1 or value == 'true':
+                    for e in list(child_elements(node)):
+                        cloned = node.cloneNode(deep = True)
+                        eval_all(cloned, macros, symbols)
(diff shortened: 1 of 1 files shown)


jonbinney:
Thanks for implementing this piyush!

davetcoleman:
I'm excited about this feature - it is a much needed one that we have actually been talking about with regards to Gazebo URDF models. Thanks!

piyushk:
No problem. I have been used a very nasty wrapper around xacro that uses
sed to get this functionality. I should have implemented this in the first
place.

davetcoleman:
Documented http://ros.org/wiki/xacro#Conditional_Blocks

jbohren:
Sweet.

PR #10, opened 2013-06-19
adding include-once attribute for preventing re-inclusion of unique tags

jbohren:
This PR makes it so adding `include-once="true"` to the root element of a xacro file will prevent that file from being included multiple times.


wjwwood:
I think this is a duplicate of https://github.com/ros/xacro/pull/11, I will close it in favor of the new rebased version which will enable continuous integration with travis.

PR #11, opened 2013-06-28
adding include-once attribute for preventing re-inclusion of unique tags

jbohren:
This PR makes it so adding `include-once="true"` to the root element of a xacro file will prevent that file from being included multiple times.

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -242,24 +242,30 @@ def process_includes(doc, base_dir):
                     raise XacroException("included file \"%s\" could not be opened: %s" % (filename, str(e)))
                 try:
                     global all_includes
-                    all_includes.append(filename)
                     included = parse(f)
+
+                    # Check for inclusion guard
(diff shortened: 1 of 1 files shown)


jbohren:
Bump.

mikepurvis:
I can picture this being useful for baking up a URDF with a lot of optional components having dependencies. Eg, a base platform that has a mast with several optional peripherals on the mast—including any of those peripherals brings in the mast, but if you want more than one of them, you still only want one instance of the mast. Is this an intended use-case, or are there others you have in mind?

Multiple include is normally a problem because of issues with redefinition, but xacro doesn't mind you redefining a macro.

wjwwood:
Can you rebase this pull request?

jbohren:
> Can you rebase this pull request?

Yeah, I will.

wjwwood:
@jbohren thanks, I guess you are going to open a new pull request then instead of updating this one?

jbohren:
> @jbohren thanks, I guess you are going to open a new pull request then instead of updating this one?

I'll update this one, but I'll leave it closed until it's ready for review again.

wjwwood:
Ok.

PR #18, opened 2014-01-31
Globbing for <xacro:include>

mikepurvis:
In addition to:

```
<xacro:include filename="$(find my_package)/other_file.xacro" />
```

There would be support for:

```
<xacro:include filename="/path/to/my/extras/*.xacro" />
<xacro:include filename="/path/to/my/extras/*" />
```

My interest is in using xacro to build up `launch` and `urdf`s at launch-time from installed packages, based on a platform configuration that is not known when the packages are built. So the launch and urdf are constructed using globs to mix-in additional components that correspond to accessories configured on our vehicles.

At present, we achieve the desired effect using [mklaunch](https://github.com/clearpathrobotics/robot_upstart/blob/hydro-devel/scripts/mklaunch) and [mkxacro](https://github.com/clearpathrobotics/robot_upstart/blob/hydro-devel/scripts/mkxacro) scripts in the robot_upstart package, but I believe this functionality would be best provided by xacro—would be happy to prepare a PR for indigo.

Thoughts?


mikepurvis:
Turned out to be a fairly simple implementation, so I just went ahead with it—the main consequence of this approach is that including of files with [, *, or ? in the filename breaks, since these will attempt to be expanded. The `filename=""` attribute is only expanded when it contains [, *, or ?.

I can rework it if there'd be a preference to something more explicit, eg, a `glob="yes"` attribute on the include tag, or a new tag altogether. This way is simple and straightforward, though. 

In any case, I've added three new tests which check the classical behaviour, the globbing behaviour, and the file-not-found scenario.

Pinging @wjwwood, @jbohren, @jonbinney for review.

src/xacro/__init__.py
```diff
@@ -230,36 +233,44 @@ def process_includes(doc, base_dir):
...
+            for filename in filenames:
+                global all_includes
+                all_includes.append(filename)
                 try:
-                    f = open(filename)
+                    with open(filename) as f:
+                        try:
+                            included = parse(f)
+                            print(f)
+                            print(included)
```
wjwwood: Are these print statements needed?

wjwwood:
This looks ok, I need to run it locally, but the tests seem valid.

Separately I need to make a pull request to clean up and make Python3 ready this code.

src/xacro/__init__.py
mikepurvis: Oh blah—no, that's lazy debugging slipping through, thanks.

mikepurvis:
Okay, removed the two `print` calls introduced by me, and one leftover from previously in the other exception handler.

It'd probably be fine to just release this into Hydro, but if you'd prefer to hold it for indigo, you could merge it in there, and then make the Python3 changes in that branch.

wjwwood:
I made a pull request for the code clean up:

https://github.com/ros/xacro/pull/19

wjwwood:
Once I merge that pr I'd like you to rebase this one.

mikepurvis:
Okay, sounds good.

jbohren:
While this tag is on topic, any interest in #11?

wjwwood:
@jbohren I will look at it for you. Once #19 is merged, I'll get you to rebase #11 as well.

mikepurvis:
Rebase done. I can't (easily) test python 3 due to the issues in ros_comm, but I believe I applied the same changes you made (using except-as, limiting line lengths)

wjwwood:
I tested it locally, looks OK:

```
william@dosa ~/devel/xacro (git:mikepurvis-hydro-devel:3754c07)
% nosetests-3.3 -s test/test_xacro.py
...............
----------------------------------------------------------------------
Ran 15 tests in 0.044s

OK

william@dosa ~/devel/xacro (git:mikepurvis-hydro-devel:3754c07)
% nosetests-2.7 -s test/test_xacro.py
...............
----------------------------------------------------------------------
Ran 15 tests in 0.026s

OK
```

PR #21, opened 2014-02-18
Add roslint target to xacro; two whitespace fixes so that it passes.

mikepurvis:
Not connected to the tests at this point—you have to manually `catkin_make roslint` or `catkin_make roslint_xacro` to see the linter output.


mikepurvis:
Note that you'll need to grab roslint from source, or use the deb which should shortly be in shadow for hydro and indigo.

PR #24, opened 2014-03-14
backported if statements from hydro-devel 

guihomework:
If statement support are really interesting and worth sharing with users constrained to stick to groovy for some more time.

backported if statements from hydro-devel and improved float evaluation for unless also.
backported logic statements test including unless.

### DIFF
--- a/src/xacro.py
+++ b/src/xacro.py
@@ -504,6 +504,22 @@ def eval_all(root, macros, symbols):
                     raise XacroException("Block \"%s\" was never declared" % name)
 
                 node = None
+            elif node.tagName in ['if', 'xacro:if', 'unless', 'xacro:unless']:
+                value = eval_text(node.getAttribute('value'), symbols)
+                try: 
+                    if value == 'true': keep = True
+                    elif value == 'false': keep = False
+                    else: keep = int(float(value))
+                except ValueError:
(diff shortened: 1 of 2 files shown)


wjwwood:
I just added proper CI for groovy in this pull request: https://github.com/ros/xacro/pull/26

Can you rebase your pull request so that CI will work correctly?

guihomework:
I am new to git and I suppose I managed to rebase on your pull request although with undesired intermediate commits that could certainly be avoided with fancier commands I am not able to use yet.

wjwwood:
I can probably help with that:

```
$ git remote add upstream https://github.com/ros/xacro.git
$ git fetch upstream
$ git checkout groovy-devel
$ git reset --hard upstream/groovy-devel
$ git cherry-pick 2142b40
$ git push -f origin groovy-devel
```

The above assumes your `origin` is your fork of this repository.

guihomework:
I had mostly done what you suggested in fact but then my push was done without the -f and failed. And for this reason I had to merge and rebase and such. the -f solves everything but is unrecommended from what I read, even more when the repository is public (of course nobody uses the repository so it is less important). I thought it would confuse also github and the pull request if I had just lost the part of the history but apparently this worked nice.

wjwwood:
Thanks.

Yeah, force is required when you are changing the history of a branch, which normally is bad because everyone who has a local checkout of the branch will be out of sync then, but for a pull request I think it is ok because you want to clean up and hide the work that went on in the devel branch before merging it into the master.

PR #42, opened 2014-12-08
Allow xacro macros to have default values for parameters

paulbovbel:
Allow xacro macros to have parameters defined in the style:

`<xacro:macro name="test_macro" params="param1:=default_value param2 *block_param">`

and then allow the usage:

`<xacro:test_macro param2="test2">`

Where param1 receives default_value if not specified in the macro usage.


paulbovbel:
Allows for more complex macro'ing with reduced verbosity

paulbovbel:
Tests added

PR #43, opened 2014-12-08
fixes #23, specify arg defaults

### LINKED ISSUE #23: No way to specify arg defaults
Not sure how important this is, but it would be nice to have the &lt;arg name="foo" default="123" /> syntax from roslaunch.

paulbovbel:
Fix #23, for substitution in `$(arg foo)`, this pull allows default arg syntax:

`<xacro:arg name="foo" default="123" />`

If neither default nor command line arg is present, xacro will throw an error as before.


paulbovbel:
Tests added

paulbovbel:
Hi @codebot9000 , would you be amenable to merging this and https://github.com/ros/xacro/pull/42? I can update the doc on roswiki once you release as well.

PR #46, opened 2015-01-27
Include CATKIN_ENV params at build time.

mikeodr:
`CATKIN_ENV` param is not included at build time and causes the `$(find pkg)`
macro to fail if the package is local to the build space and not found in `ROS_PACKAGE_PATH`

`CATKIN_ENV` param has been added to the `COMMAND` call in process execution
makefile argument.


mikeodr:
Yeah I was concerned about the travis failure until I dug into it.
Will ignore for now.

PR #47, opened 2015-01-27
travis-ci: use catkin_make

wjwwood:
@codebot9000 Let's iterate here until it passes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 python:
-  - "2.7"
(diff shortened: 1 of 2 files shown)


wjwwood:
The remaining problem is this:
- https://github.com/ros/catkin/pull/628
- https://github.com/ros/catkin/issues/626

And we can't solve it correctly right now. I'll try the workaround.

wjwwood:
@codebot9000 for your review.

Issue #49, opened 2015-02-15
Feature request: index support into bundled params


freyzor:
Trying to write modular xacto definitions I find myself really wanting to be able to index bundled xyz or rpi params passed along into macros.  This would greatly simplify making components.

To give an example here is a bioloid F2 bracket part used to compose composit links.

``` xml
    <macro name="geometry_F2">
        <geometry>
            <mesh filename="package://${PACKAGE}/meshes/F2.stl" scale="${M_SCALE_3}"/>
        </geometry>
    </macro>
    <macro name="visual_F2" params="xyz rpy">
        <visual>
            <origin xyz="${xyz}" rpy="${rpy}" />
            <geometry_F2 />
            <color_grey60 />
        </visual>
    </macro>
    <macro name="collision_F2" params="xyz rpy">
        <collision_box xyz="${xyz}" rpy="${rpy}" size="0.025 0.0485 0.0375" />
    </macro>
    <macro name="part_F2" params="xyz rpy">
        <collision_F2 xyz="${xyz}" rpy="${rpy}" />
        <visual_F2 xyz="${xyz}" rpy="${rpy}" />
    </macro>
```

I then use it in to create a link like so

``` xml
    <macro name="coxa_bone" params="parent name">
        <link name="${name}_link">
            <inertia_default />
            <part_ax12 xyz="0 0 0" rpy="0 0 ${MPI}" />
            <part_F3 xyz="0 0 -0.0415" rpy="0 ${MPI} 0 "/>
            <part_F2 xyz="0 0 -${L_COXA}" rpy="0 0 ${MPI2} "/>
        </link> 
        <joint_revolute name="${name}" parent="${parent}" xyz="0 0 0" rpy="0 0 0" 
            vel="100" llimit="${MIN_RAD_COXA}" ulimit="${MAX_RAD_COXA}" axis="0 1 0" />   
    </macro>
```

If I want to provide an offset to the xyz in the collision_F2 macro as it does not align to the stl mesh I'm unable to extract the x y z components again to add an offset.  You can get around this by passing in the values separately but that does brake the encapsulation and makes the whole thing pretty verbose again. 

what I'd like to be able to do would be something like the following.

``` xml
    <macro name="collision_F2" params="xyz rpy">
        <collision_box xyz="${xyz[0]+0.03 xyz[1]-0.4 xyz[3]}" rpy="${rpy}" size="0.025 0.0485 0.0375" />
    </macro>
```

This way I can keep the implementation detail neatly contained.

[...]

PR #50, opened 2015-02-17
various improvements to xacro

rhaschke:
Dear maintainer,

this isn't a classical pull request, but rather a <b>pull proposal</b>. It comprises:
- fixup of unittests / `xml_matches()`, handling of non-element nodes in `<include>`, `<if>`, `<macro>` as  indicated in previous [pull request #48](https://github.com/ros/xacro/pull/48)
- dynamic macro names: `<xacro:call macro="${variable}"/>`
- XML processing in read-order to avoid issues with multiply defined macros or properties
- python-based evaluation of expressions: utilizing `eval()` instead of the hand-coded parser

All changes build on top of each other to some extend. They have associated branches in [our xacro fork](https://github.com/ubi-agni/xacro/tree/agni). A more detailed motivation for the individual changes can found in the [README](https://github.com/ubi-agni/xacro/tree/agni#xacro)

I'm aware of the fact, that these changes heavily modify the original code. To get an overview of all changes you should follow the individual commits, which I carefully packaged into meaningful and easily comprehensible steps. Of course, I'm open for questions and discussion as well ;-)

I'm also aware of the fact, that my changes will have some impact on existing xacro files out there. I see namely three changes xacro users need to take care of:
- Ordered processing will break stuff, that relied on the specific manner xacro was processing includes, macros and properties - utilizing the latest definition only for evaluation. However, even the rather complex `pr2 example` doesn't rely on this fact, but successfully compiles in both modes.
  Because I'm aware of this fact, I decided to enable the ordered-processing optionally only, with the command line option --inorder
- copying any type of top-level XML nodes from `<include>`, `<if>`, `<macro>` will also change the previous behaviour. This will be irrelevant however for urdf, because urdf only considers element nodes anyway.
- Finally, the python-based evaluation will modify the handling of integer division: While the old parser silently switches to floats when doing integer division, python eval - as any other programming language - distinguishes between integer and float division. From my point of view, this will be the most important change to take care of by end users.

By the way, why did you opted for an own, hand-written expression parser instead of using python's capabilities? I don't see any benefit for this.

Having said all this, I would be happy if you are interested in some of these contributions and would like to consider them for the next or over-next ROS release.

Kind regards, Robert

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -39,6 +39,7 @@ import re
 import string
 import sys
 import xml
+import ast
 
 from xml.dom.minidom import parse
 
@@ -129,9 +130,7 @@ class Table:
         if isinstance(value, basestring):
             try:
                 # try to evaluate as literal, e.g. number, boolean, etc.
-                value = eval(value)
-                # TODO: store integers as float to maintain backward compatibility?
-                # if isinstance(value, int): value=float(value)
+                value = ast.literal_eval(value)
             except:
                 # otherwise simply store as original string for later (re)evaluation
                 pass
@@ -557,11 +556,14 @@ def eval_all(root, macros={}, symbols=Table()):
                 node = None
 
             elif node.tagName in ['if', 'xacro:if', 'unless', 'xacro:unless']:
-                value = str(eval_text(node.getAttribute('value'), symbols))
+                value = eval_text(node.getAttribute('value'), symbols)
                 try: 
                     # try to interpret value as boolean
-                    # dict explicitly enables 'true' and 'false'
-                    keep = bool(eval(value, dict(true=True, false=False)))
+                    if isinstance(value, basestring): 
+                        if   value == "true": keep = True
+                        elif value == "false": keep = False
+                        else: keep = ast.literal_eval(value)
+                    else: keep = bool(value)
                 except:
                     print ("if failure", value, type(value))
                     raise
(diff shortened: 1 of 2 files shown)


rhaschke:
Nice to hear that you like the suggestions. Regarding integer division, I would like to look into some python capabilities first. As the following links suggest, python itself is struggling with integer vs. float division since version 2.2 too and they suggested the operator // (double slash) for integer division. In python 3 single slash will default to float division. This new behaviour can be enforced in python >2.2
doing a "from **future** import division". This could be an option to test, but I will need to find some time to do it.
http://python3porting.com/preparing.html
https://www.python.org/dev/peps/pep-0238/

Introducing another variable indicator #{} is also an option, but it is not very appealing.

rhaschke:
At least, in an ipython shell, the new division behavior is propagated to eval(). However, I'm not sure what side effects that might have in xacro itself. But probably xacro doesn't use calculations by itself.
Also, I'm not sure how to realize version-agnostic python code.

rhaschke:
I added the `from future import division` statement and it works on the unittests! No sideeffects observed so far. I have reverted the changes to the pr2 files accordingly.
freyzor also pointed out an [evaluation issue](https://github.com/ros/xacro/issues/49#issuecomment-75138763), which I could easily fix. The updates are committed to [our fork](https://github.com/ubi-agni/xacro/tree/python_eval). 
While searching for a solution to this issue, I stumbled upon some security warnings regarding eval():
http://nedbatchelder.com/blog/201206/eval_really_is_dangerous.html

Obviously, if there is bad code entered in evaluated expressions, hackers can do nasty things...
I considered this issue to some extend in my first attempt, not providing access to the globals() dictionary. However, I wasn't aware, that one can still import stuff. I added the simple security measure pointed out on the mentioned site to forbid access to `__builtins__`.

rhaschke:
Done.

PR #59, opened 2015-04-01
Roslaunch <arg> xml tag being removed

mikeodr:
If using launch files generated by xacro, the `<arg>` tag was being removed.

Added a unit test to catch this condition and removed `arg` from the `elif` check.
The check now only removes `xacro:arg`.

I've checked this over with the author of pull #42.


mikeodr:
@paulbovbel FYI.

mikeodr:
Any issue with this pull request being against indigo? I'll cherry pick this into jade as well once this is complete?

mikeodr:
@codebot Shall I take care of the cherry pick to jade? Or will that be done with a merge otherwise in updating jade?

Issue #57, opened 2015-03-16
increase speed by replacing xml.minidom with lxml.etree


codebot:
It appears that we can speed up xacro a lot by moving to lxml.etree instead of using xml.minidom. Obviously this would be a big change.

@mikepurvis , @paulbovbel , @mikeferguson , @wjwwood , and anyone else: any comments or feedback on this proposal?

Related: I'm also interested in adding some more large / complex "full-robot" test cases to xacro, to complement to the PR2 full-robot test case. I think we'd have more confidence when making big changes like this to xacro if we had more large test cases. Anyone have suggestions for which robots to add?

mikepurvis:
Flagging @TheDash, who is presently maintaining PR2.

Is the speed an issue for anyone? Our xacro is all extremely simple, and the compositing step only runs once on launch anyway.

If there are resources available on this, I would be interested in seeing a survey of the many, many other schemes which now exist for templating and compositing XML, and an assessment given of the cost of creating a ROS wrapper/plugin for one of them which gives the necessary path-resolution stuff but otherwise weakens the coupling to the ROS ecosystem. Something like Genshi or PyMeld, or even non-XML tools like Empy or Jinja2.

codebot:
@rhaschke has reported that some of their large files are taking up to 30 seconds.

UltronDestroyer:
Never had an issue with the xacro compiling. Not speed at least, but with other things (parameters, tests, structure and composite capability)

UltronDestroyer:
If needed, we can run Vtune on Xacro to see where some bottlenecks in the code are.

codebot:
@rhaschke  has done some profiling, which he summarized briefly in #56 , which is why he suggested we switch to a faster XML parser.

mikeferguson:
I've not really had an issue with speed -- our use of Xacro is pretty simple. Most of our robot is a single URDF (this is going to be more and more common as more people export their URDF directly from Solidworks) and then the gripper gets added via Xacro (since it is a modularity point). My other use of Xacro is to incorporate the tags required for Gazebo.

I can supply test Xacros in the future, but not at this time as the robot is not yet released.

wjwwood:
In principal the result should be the same, but there is a lot of opportunity to introduce new bugs. It could be possible to make it an option, depending on how complex the changes are. It might be a good opportunity to introduce an abstract internal to xacro, allowing it to use different xml engines.

[...]

PR #62, opened 2015-04-23
further improvements: merged some recent pull requests (#60, #66, #68)

rhaschke:
Sorry for the long silence. My priorities were shifted...
Here is the promised (#56) pull request to improve QuickLexer performance.
It essentially instantiates the QuickLexer's regexps only once and reuses them.

I stopped working for now on #57. There are more pressing issues ;-)

### DIFF
--- a/cmake/xacro-extras.cmake.em
+++ b/cmake/xacro-extras.cmake.em
@@ -9,8 +9,20 @@ set(_xacro_py
 @[end if]@
 
 macro(xacro_add_xacro_file input output)
+  set(options OPTIONAL INORDER)
+  set(multiValueArgs REMAP)
+  cmake_parse_arguments(_XACRO "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
+  if(_XACRO_INORDER)
+    set(_XACRO_INORDER "--inorder")
+  else()
+    unset(_XACRO_INORDER)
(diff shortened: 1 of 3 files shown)


rhaschke:
The second commit introduces some basic python computing capabilities to xacro property handling.
Particularly, some basic data types are added to allow for string manipulation (split, join, append, etc.) as well as explicit type conversion (to int or float).
Further, I added the map() function to allow some basic looping operations of lists. 

This allows e.g. to parse a string vector into a list of floats: map(float, "0 1 2".split())
or to join a list of values into a string: " ".join(map(str, [0, 1, 2]))

rhaschke:
@codebot 
There was a long period of silence from your side regarding xacro. 
Meanwhile several pull requests have been proposed from different people leading to a divergence of the code base. I took the effort to merge most of them (#60, #66, #68). Pull request #67 can be dropped, because the present `--inorder` processing code branch is a more elegant way to handle issue #63 (as a test case proves).

Regarding #65, which is very nice too, I suggest to strip down scripts/xacro (similar to scripts/rostopic)
and do the cli processing threre. xacro.main() would then take cli arguments as python function args and could be called from the tests too. 
I already implemented the first part (stripping down scripts/xacro) and could continue with merging #65.

What do you think?

PR #68, opened 2015-05-11
Fixed a bug that caused property values not to be converted to Python…

peci1:
… types when assigned from an <arg> tag.


rhaschke:
Valid patch. I considered it for my [fork](http://github.com/agni/xacro).

PR #69, opened 2015-05-21
fix encode bug

xuhao1:
Line 704 may cause some problem deal with utf-8, it always cause bugs when run SITL of PX4, so I change the encode to "utf-8" and it will fix these bugs

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -701,5 +701,5 @@ def main():
     for comment in banner:
         doc.insertBefore(comment, first)
 
-    open_output(output_filename).write(doc.toprettyxml(indent='  '))
+    open_output(output_filename).write(doc.toprettyxml(indent='  ').encode("utf-8:))
     print()


awesomebytes:
The line:
https://github.com/ros/xacro/pull/69/files#diff-f2b480897d82b8d4fc9cedec71732303R704

Says:

```
.encode("utf-8:))
```

But probably you mean:

```
.encode("utf-8"))
```

(Although with my URDF that I cannot currently post I get another error now:

```
  File "/home/sampfeiffer/tmp/xacros/src/xacro/scripts/xacro", line 53, in <module>
    xacro.main()
  File "<string>", line 693, in main
  File "<string>", line 616, in eval_self_contained
  File "<string>", line 523, in eval_all
  File "<string>", line 523, in eval_all
  File "<string>", line 588, in eval_all
  File "<string>", line 414, in eval_text
  File "<string>", line 397, in handle_expr
  File "<string>", line 1, in <module>
TypeError: can't multiply sequence by non-int of type 'float'
```

)

On Ubuntu 14.04 with Indigo we are experiencing the same issue of:
`UnicodeEncodeError: 'ascii' codec can't encode character u'\xba' in position 23116: ordinal not in range(128)`

And I cannot manage to find the offending character (it should be º but I cannot find it doing greps).

awesomebytes:
As a workaround we did a checkout of the xacro package of hydro (branch hydro-devel) and just used that in our workspace.

PR #71, opened 2015-05-22
deprecate xacro.py

rhaschke:
This pull request suggests to deprecate the `xacro.py` script in favor of `xacro`. 
It is designed to be applied **on top of #70** (otherwise unittests fail).

The first commit cleans up `scripts/xacro`, which is the primary binary to call for xacro, and removes `xacro.py`. This top-level script should essentially only call `xacro.main()`. However, it contained a lot of code to filter-out the current path from `sys.path`. (This probably will fail, if one runs xacro accidentally in the required xacro module path.) So, I stripped the code down to the essentials.

I suggest to remove the script `xacro.py` completely (at least in the long run). 
Essentially, the `sys.path` filtering is only necessary because `xacro.py` will be mistaken as the `xacro` module. The standard ROS way to name python binaries seems to go without `.py` ending (c.f. [rostopic](https://github.com/ros/ros_comm/blob/indigo-devel/tools/rostopic/scripts/rostopic)). 

Of course, xacro.py is heavily in use currently (probably because it's heavily refered to in the [docs](http://wiki.ros.org/urdf/Tutorials/Using%20Xacro%20to%20Clean%20Up%20a%20URDF%20File#Using_Xacro)). Hence, the second commit re-adds xacro.py as `scripts/legacy-xacro.py`, installing to `share/xacro/xacro.py` as before. However, using this `xacro.py` issues a deprecation warning now to allow a smoother transition.

Do you think, one should trigger this transition?

If so, there is one regression to mention: By placing xacro.py into `scripts` it will not be available from a local devel space. However, I thought placing it into scripts would be more appropriate. Most people will use an installed version anyway (and get the deprecation warning then).

P.S.: The two commits are separated by purpose to facilitate later removal of `xacro.py`.
P.P.S: Not applying the patch on top of #70, will result in unittest errors, because xacro.py will not be found in the old location. However, #70 removed the explicit search for a `xacro.py` anyway.


scripts/legacy-xacro.py
```diff
@@ -56,5 +56,6 @@
 # Restore the path
 sys.path = backup_path
 
-if __name__ == '__main__':
-    xacro.main()
+xacro.main()
+
+print('Use of deprecated xacro.py. Please, resort to xacro.', file=sys.stderr)
```
codebot: Can you change this message to something simpler, like "xacro.py is deprecated; please use xacro instead."

scripts/xacro
```diff
@@ -29,25 +29,5 @@
...
-sys.path = filter(lambda a: a not in [this_dir, this_dir_cwd], sys.path)
-
 import xacro
-
-# Restore the path
-sys.path = backup_path
-
-if __name__ == '__main__':
-    xacro.main()
+xacro.main()
```
codebot, line 33: this is awesome

rhaschke:
Simplified deprecation message as suggested.

CMakeLists.txt
```diff
@@ -9,12 +9,19 @@ catkin_package(
...
+
+catkin_install_python(PROGRAMS scripts/xacro
+  DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION})
 
-install(PROGRAMS scripts/xacro DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION})
+## For backwards compatibility we also install as xacro.py
+## install into share/xacro (instead of lib/xacro)!
+install(PROGRAMS scripts/legacy-xacro.py
+  DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}
+  RENAME xacro.py)
```
codebot, line 24: This is a tough call. I agree that it feels more correct to move the deprecated xacro.py into scripts/xacro.py, but ... I suspect that a decent number of people are using xacro from catkin devel spaces (e.g., to use new xacro-jade features in Indigo). So, I'm thinking that we should leave xacro.py in its current location for Jade, since it's been there for many releases that I'm afraid to both deprecate it and move it simultaneously. This will be printing a warning every time (can we have it print a warning in yellow/red? not sure how hard that is), so hopefully people will gradually migrate their code to clean up the warnings. Then, in K-turtle we can move xacro.py to scripts/xacro.py , or perhaps even remove it completely. What do you think?

rhaschke:
I was kind of undecided on this. Both has its pros and cons.
However, I follow your argumentation and will move xacro.py back to its original location.
I can look into some coloring too.

scripts/legacy-xacro.py
```diff
@@ -56,5 +56,6 @@
 # Restore the path
 sys.path = backup_path
 
-if __name__ == '__main__':
-    xacro.main()
+xacro.main()
+
+print('xacro.py is deprecated, please use xacro instead.', file=sys.stderr)
```
codebot: Sorry to request another tweak to this message, but I think a lot of people will see this. I debated suggesting to add the \033[93m type of color escape sequence, but that will look weird on non-color terminals. How about we just add WARNING: in front of the error message, so it reads "WARNING: xacro.py is deprecated; please use xacro instead"  Thanks!

rhaschke:
I prepared everything.

rhaschke:
I forgot to mention: Coloring is used if the terminal provides color. Otherwise "WARNING: " is prepended to the deprecation message.

PR #77, opened 2015-05-24
improved xacro's cmake macros

rhaschke:
- replaced macro with function: have local variable namespace in function!
- xacro_add_xacro_file() automatically determines output file from input (removing .xacro suffix).
  If that fails, a fatal error is raised. The created output file is returned to caller.
- added xacro_install() to allow installation into both, devel and install space.
- replaced conveniency macro xacro_add_target() with function xacro_add_files()
- added docs of functions to source file

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,8 @@ install(PROGRAMS xacro.py DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
 if(CATKIN_ENABLE_TESTING)
   ## Add folders to be run by python nosetests
   catkin_add_nosetests(test)
+  ## add tests for xacro's cmake functions
+  add_subdirectory(test)
 endif()
 
 roslint_python()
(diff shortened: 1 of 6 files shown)


rhaschke:
The second commit adds a basic unittest framework for xacro's cmake macros.
The shell script [`test/test-cmake.sh`](https://github.com/ros/xacro/blob/62a2f8e07320fc42fc78f7c431693ccdc7aeb95a/test/test-cmake.sh) simply runs the cmake build chain and the [`test/test-xacro-cmake/CMakeLists.txt`](https://github.com/ros/xacro/blob/62a2f8e07320fc42fc78f7c431693ccdc7aeb95a/test/test-xacro-cmake/CMakeLists.txt) calls some xacro macros. Currently only the build chain is tested. Of course, one could check for specific results too.

rhaschke:
Probably I should prepend the PACKAGE_NAME to every custom cmake target. Otherwise multiple projects using the same target name (e.g. xacro) would conflict too...
I didn't noticed that issue with the new [catkin tools](https://catkin-tools.readthedocs.org/en/latest/), which probably build each package individually, while catkin_make merges everything into a single cmake config?

rhaschke:
There are some other places I need to take care of.

rhaschke:
Fixed all locations now.

PR #84, opened 2015-06-08
improved error reporting

rhaschke:
This is an attempt to improve error reporting to the normal xacro user. So far it was rather difficult to figure out, which piece of code (particularly which file) caused a certain error. With the proposed patches, xacro keeps track of its processed file stack and the function print_location() can be used to report it at any place. (With #75 I added recursive file reporting already, but that wasn't flexible enough to report the file stack anytime). 

The PR partially reverts #75 in the sense, that the `filename` argument that was passed through a lot of functions, is removed again. Instead, there is a global variable `filestack` where push/pop a newly processed file onto/from. This allows to store the filestack together with a macro definition allowing to know where a macro was defined when it will fail later during instantiation.

Further, I enabled XacroException to wrap another exception and augment its error message with some other information prepended or appended (which was used in the last three commits).

As you might have noted already, this commit builds on (and includes #82). Due to some overlapping changes, it wasn't possible to file them as separate PRs.

@codebot: I have some more feature extensions in the pipeline, namely yaml-import (of dict properties),
dynamically named element tags (through renaming), better handling of XML comments, and namespacing for includes. Please, give me a shout when you have some time to review them.

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -351,7 +351,7 @@ def get_include_files(elt, parent_filename, symbols):
         # Globbing behaviour
         filenames = sorted(glob.glob(filename_spec))
         if len(filenames) == 0:
-            print(include_no_matches_msg.format(filename_spec), file=sys.stderr)
+            warning(include_no_matches_msg.format(filename_spec))
     else:
         # Default behaviour
         filenames = [filename_spec]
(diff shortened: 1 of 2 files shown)


rhaschke:
The commits are not listed in the correct order on github - probably, because I cleaned them up and reordered them by interactive rebasing. To be sure, please first handle #82 and I will rebase onto that.

PR #88, opened 2015-06-09
yaml support

rhaschke:
I added `yaml` support to load a `dict` property from yaml. 
Having more complex, hierarchical properties turned out to be very useful in our setup where we compose a `robot_description` from some basic scaffold attaching arms and tools as dynamically specified on the command line. The properties of those robot components are summarized in a yaml file which can now be loaded into `xacro` and used for evaluation. Syntax is as easy as:
`${load_yaml('file.yaml')}`

A unittest illustrates basic usage.

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -77,6 +77,34 @@ def restore_filestack(oldstack):
     filestack = oldstack
 
 
+def abs_filename_spec(filename_spec):
+    """
+    Prepend the dirname of the currently processed file
+    if filename_spec is not yet absolute
+    """
+    if not os.path.isabs(filename_spec):
+        parent_filename = filestack[-1]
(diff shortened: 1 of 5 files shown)


paulbovbel:
This is fantastic. Is your setup available publicly?

rhaschke:
Our setup description is not (yet) publicly available.
However, there is a small usage example in the unittests: test_yaml_support() using [settings.yaml](https://github.com/ros/xacro/blob/jade-devel/test/settings.yaml)

`````` xml
  <xacro:property name="settings" value="${load_yaml('settings.yaml')}"/>
  <xacro:property name="type" value="$(arg type)"/>
  <xacro:include filename="${settings['arms'][type]['file']}"/>
  <xacro:call macro="${settings['arms'][type]['macro']}"/>
</a>```
``````

PR #90, opened 2015-06-10
improved processing

rhaschke:
### improved processing
- recursive (instead of iterative) `eval_all()`
- reusable `process_include()`
- `replace_node()` function to replace a xacro tag by some other content (single node, list of nodes, None)
- avoid reprocessing of nodes
- avoid deep copy whereever possible (speedup)

Rewriting eval_all() as a recursive function allows for a lot more code reusability. Previously, eval_all would traverse beyond the last child node if siblings or a parent was available. This required to deep-copy nodes to dissociate them from the DOM tree, although copying wasn't necessary semantically.

The new function `replace_node()` centralizes the task of replacing a xacro element by some other content, which was repeatedly hand-coded before.
### better comment handling

With #48 I pull-requested some changes to consider all node types for the final XML (that is also comments within macros). However, one clearly wants to distinguish between xacro-only comments and comments that should go to the final document. 
Now I came up with the following solution: All comments that are directly in front of a `xacro` element (eventually separated by some white-space-only text) will be considered as xacro-related and removed.
All other comments (and comments separated by more than one newline from the next one) will be considered as desired and will go to the final XML document. This way one has fine-grained control over the comments. Here is an example:

``` xml
<!-- Kept comment, because in front of a non-xacro-tag -->
<my_tag/>
<!-- Kept as well, because separated by a newline from the next comment -->

<!-- This comment will be removed -->
<!-- This one too, because they are in front of a xacro tag like macro, property, if, arg, include, ... -->
<xacro:*/>
```

Obviously, I had to undo the adaptions to the `pr2` gold standard introduced in #48.
A new unittest test_ignore_xacro_comments() checks the desired behavior.

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -314,51 +314,39 @@ class QuickLexer(object):
 
 def first_child_element(elt):
     c = elt.firstChild
-    while c:
-        if c.nodeType == xml.dom.Node.ELEMENT_NODE:
-            return c
+    while c and c.nodeType != xml.dom.Node.ELEMENT_NODE:
         c = c.nextSibling
-    return None
+    return c
 
 
-def next_sibling_element(elt):
-    c = elt.nextSibling
-    while c:
-        if c.nodeType == xml.dom.Node.ELEMENT_NODE:
-            return c
+def next_sibling_element(node):
+    c = node.nextSibling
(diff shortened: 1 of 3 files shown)


rhaschke:
With this PR, the speed improvement wasn't yet dramatic. Patching [xml.dom.minidom](https://bugs.python.org/issue24424) gave me a **factor of 10** (on a huge file).

PR #92, opened 2015-06-10
allow renaming of element names

rhaschke:
Recently I wanted to define a macro that should create a XML fragment with dynamic tag names,
i.e. tag names generated from macro parameters. Unfortunately, this wasn't yet handled by xacro.
This is a simple patch to allow it. Do you like the `<xacro:rename>` syntax?

``` xml
<xacro:rename xacro:name="<new element name>" att="val" ...>
   ... content ... 
</xacro:rename>
```

By purpose, I've put the name attribute into the xacro namespace as well - to allow for a native name attribute within the generated element too.

Having xacro extended towards dynamic tag name generation, there is still one missing feature: Generating dynamic attribute names. However, I don't yet have a nice syntax for it yet. One could think of

`````` xml
<xacro:attribute name="" value=""/>```
``````

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -644,6 +644,17 @@ def eval_all(node, macros, symbols):
                 remove_previous_comments(node)
                 replace_node(node, by=None)
 
+            elif node.tagName == 'xacro:rename' \
+                    and check_deprecated_tag(node.tagName):
+                name = node.getAttribute('xacro:name')
+                if name is None:
+                    raise XacroException("xacro:rename: 'xacro:name' attribute missing")
+                else:
+                    node.removeAttribute('xacro:name')
+
+                node.nodeName = node.tagName = eval_text(name, symbols)
+                continue  # re-process the node with new tagName
+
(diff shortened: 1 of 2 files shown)


wjwwood:
I've been noticing a lot of changes to xacro lately (which is great!), but we should be cautious to not forget to update documentation along the way, especially when introducing new features like this. Just a friendly reminder :smile:.

rhaschke:
Thank you for the reminder. It's on my todo list already. Most of the accepted features are already documented on http://wiki.ros.org/xacro. Have a look!

rhaschke:
That's the point. The syntax is not yet straightforward.
The idea was to add the given attribute (name=value pair) to the parent node.
Using the previous or next XML tag wouldn't allow for multiple xacro:attribute/ tags obviously.

rhaschke:
That's indeed clumsy syntax. I guess, this will be a rarely used feature (it wasn't even requested so far).
Hence, having a simple syntax (xacro:attribute) that is nicely explained on the wiki should do the job, shouldn't it?

rhaschke:
Yeah, I can rename "rename" to "element".

PR #93, opened 2015-06-10
allow namespacing for xacro:include's

rhaschke:
Composing more complex xacro documents from several, typically 3rd party sources, may cause name clashes. To avoid this, I suggest to allow for namespacing of macros and properties defined within an included file.
Employing python evaluation not only for properties, but also for macro names, I could quite easily add this feature.

However, this comes with a drawback (as before with properties): Valid macro (and property) names must be valid python identifiers. Particularly dashes, points, etc are not allowed anymore: `this-is-invalid`.

For now, macro resolving uses the old strategy first and resorts to python evaluation on failure only - keeping existing files valid. For invalid properties an exception is raised, forcing people to change names.
(For properties the risk of invalid names is smaller, because minus signs were interpreted as such before too.)

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -41,6 +41,7 @@ import xml
 import ast
 import math
 
+from keyword import iskeyword
 from roslaunch import substitution_args
 from rosgraph.names import load_mappings, REMAP
 from optparse import OptionParser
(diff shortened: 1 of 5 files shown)


rhaschke:
Please wait before merging. I will extend the unittests.

rhaschke:
Ready for merging now.

PR #94, opened 2015-06-10
minor fixes

rhaschke:
some little fixes - see the commit messages

I would like to start a migration path towards `--inorder` processing. Maintaining both code branches becomes more and more tedious. 
I suggest to keep the old processing as a default in `Jade`, but issue a deprecation warning, pointing to the [wiki](http://wiki.ros.org/xacro#Processing_Order). In the next release --inorder processing could become default with an option providing the old behavior as fallback.

What do you think?

### DIFF
--- a/cmake/xacro-extras.cmake.em
+++ b/cmake/xacro-extras.cmake.em
@@ -26,7 +26,6 @@ add_custom_target(${PROJECT_NAME}_xacro_generated_to_devel_space_ ALL)
 ##   xacro_add_xacro_file(${xacro_file} INORDER REMAP bar:=foo foo:=bar)
 ##   list(APPEND xacro_outputs ${XACRO_OUTPUT_FILE})
 ## endforeach()
-## add_custom_target(xacro_target ALL DEPENDS ${xacro_outputs})
 ## xacro_install(xacro_target ${xacro_outputs} DESTINATION xml)
 ##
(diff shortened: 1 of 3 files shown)


rhaschke:
That's difficult to judge. There are some "features" of the old processing mode, that are not supported anymore, namely defining or overriding properties/macros after their usage. From my point of view, this is bad programming style and should be avoided anyway.
I tried to point this pitfall out on the [wiki](http://wiki.ros.org/xacro#Processing_Order) (have a look). If other pitfalls will show up in future, we should augment the wiki appropriately and explain how to avoid them.

rhaschke:
Actually, I already thought about such a feature and everything is prepared for it: At least, macros know were they were defined previously. For properties some support is possible too: They can "know" when they are used and redefined (via Table.**getitem** and **setitem**).
However, we shouldn't issue a warning message on every redefinition by default, because redefinition is a perfectly valid programming tool if used properly. I suggest to add an option --verbose to explicitly enable this feature.

rhaschke:
Do you have a proposal for an option name in K-turtle to yield the old behavior as a fallback?
Also, in Jade, people might be annoyed by the deprecation warning and will probably define an alias including the --inorder option sooner or later. Hence we should provide another option to enforce the old behavior as well. What about `--oldorder`?

PR #99, opened 2015-06-26
reworked macro resolution

rhaschke:
I thoroughly reworked macro resolution. The python-eval()-based macro resolution, I introduced to enable namespaces, heavily restricted the set of possible macro names - essentially only valid python
identifiers were allowed. 
#97 has shown, that this is a too strong restriction. Particularly, it should be possible to prefix the macro name with the `xacro:` namespace in order to avoid name conflicts with XML element names.

Hence, I replaced the eval()-based macro resolution with the function `resolve_macro()`.
Now, arbitrary macro names are possible again as long as they do not contain a dot. For backwards compatibility, dots are allowed too as long as they do not collide with namespaces.
Consequently, I removed the test function `test_invalid_macro_name()` and reduced your test function `test_macro_name_with_colon()`.

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -40,6 +40,7 @@ import ast
 import math
 
 from roslaunch import substitution_args
+from rospkg.common import ResourceNotFound
 from copy import deepcopy
 from .color import warning, error, message
 from .xmlutils import *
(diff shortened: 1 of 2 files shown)


rhaschke:
The second commit simply provides a more verbose error message in case `$(find ...)` failed.
(The default message printed only the name of the requested resource, without any error message.)

Issue #100, opened 2015-06-29
auto-forward property definitions to macro parameters?


rhaschke:
I often encounter the situation, that I need to pass existing properties to identically named macro parameters (e.g. var="${var}"). I can circumvent that by not declaring these parameters. However, as soon as they are declared, xacro expects the parameters to be explicitly provided.

I suggest to relax this strong constraint and only complain if a property is not defined at all, i.e. neither explicitly in the macro params nor implicitly in the outer scope.
This will facilitate to explicitly _declare_ all used macro parameters while avoiding to explicitly _provide_ them.

What do you think?

rhaschke:
Kind of the same effect can be achieved by using default param values equal to existing properties:

``` xml
<xacro:macro name="xacro:test" params="var:=${var}"/>
```

On the one hand, this explicitly states that the default is taken from the outer scope.
On the other hand, it prevents to define a real fallback default used when no outer scope property is defined. I suggest the following hierarchy:
- an explicitly provided argument takes highest priority
- if nothing is provided, but an outer-scope property with same name is defined, this one is (silently) taken
- otherwise, the default value (if defined) is used

If none of these cases apply, the warning about a missing argument is thrown.

guihomework:
> @guihomework
> I'm still not satisfied with the latest solution. It's too verbose. What about the following syntax:
> params="forwarded:=| forwarded_with_default:=|42"
> The pipe symbol (|) indicates either forward from outer scope or use the specified default.

The pipe symbol alone looks strange. In the minds pipe is or, which fits well to tell forward OR default
Is it possible to distinguish between ${} and $ alone ? because $ would actually make sense, you don't want to specify the name again, so $ is the place holder.
`params="forwarded:=$ forwarded_with_default:=$|42"` 
Otherwise in some language, % means replacement for an argument too (in windows batch i think). 
`params="forwarded:=% forwarded_with_default:=%|42"`

rhaschke:
I prefer the `$|fallback` syntax. That's implemented on https://github.com/ubi-agni/xacro/tree/macro_args now (with all the other alternatives in the history).
@codebot: Please decide on a syntax proposed here and #101  ;-) 
I will then rebase and squash to have a clean pull request.

codebot:
could the python environment be modified to provide a "magic" symbol called "parent" or "super" that would resolve the symbols in the parent tag's scope, or return nil if the symbol doesn't exist? Then we could do things like this in the macro itself, rather than the argument list.

``` xml
[...]

PR #101, opened 2015-06-30
adopt existing property definitions for macro args (#100)

rhaschke:
That's my proposal to handle #100.

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -451,6 +451,26 @@ def is_valid_name(name):
     return False
 
 
+_forward_matcher = re.compile(r"^forward\((.*)\)$")
+def parse_default_arg(s):
+    """
+    parse a default argument spec for macro parameters
+    particularly handle forward(variable,default)
+    :param s: argument spec string
+    :return: pair of (variable, default), where any one may be None
+    """
+    m = _forward_matcher.match(s)
+    if not m:  # simple default value
+        return (None, s)
+    args = m.group(1).split(',',1)
+    if len(args) == 1:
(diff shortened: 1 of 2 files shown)


rhaschke:
Already in our lab, this proposal is discussed very controversially. Silently adopting existing property definitions from the outer scope will impede bug tracking: One cannot explicitly recognize where a particular argument comes from. 
To make things more explicit, I suggest now another syntax: Directly in the macro definition, one could specify that properties should be adopted (or forwarded) from the outer scope. This could use the existing defaults syntax, i.e. `params="var:=${forward(var, default)}"`
The new global function forward would try to find var in the outer scope and use it. If not found, it will fallback to default (if provided) or raise an exception.

rhaschke:
Added the xacro keyword `forward` to allow for forwarding properties. Syntax is as follows:

``` xml
    <xacro:property name="var" value="42"/>
    <xacro:macro name="foo" params="arg:=forward(var,${3*(7+1)})">${arg}</xacro:macro>
    <xacro:foo/>
</a>
```

I couldn't define forward as an exposed python function, because the current scope's `symbols` are not available to the function. Hence, forward is _not_ enclosed in evaluation brackets `${}`.

rhaschke:
@guihomework
I'm still not satisfied with the latest solution. It's too verbose. What about the following syntax:
`params="forwarded:=| forwarded_with_default:=|42"`
The pipe symbol (`|`) indicates _either forward from outer scope or use the specified default_.

rhaschke:
@codebot: I will close this PR for now. We should continue the discussion for the desired solution on #100.

PR #102, opened 2015-06-30
bash completion

rhaschke:
I worked a little on a bash completion function. Besides completing the options, it also parses the xacro file (using `awk`) to list all handled `<xacro:arg>s`.
Question is: Where to install the script? I have it manually copied to /etc/bash_completion.d/xacro.

### DIFF
--- a/scripts/completion.bash
+++ b/scripts/completion.bash
@@ -1,0 +1,39 @@
+function _file_arg()
+{
+  # search backwards for an existing filename argument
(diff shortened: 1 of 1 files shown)


rhaschke:
For manual installing, I would prefer creating the symlink or copying the file to /etc/bash_completion.d
However, considering debian packages, one could think about an automated install. But, I don't know where this would go...

PR #103, opened 2015-07-01
issue warning when attempting to overwrite existing global property

rhaschke:
warn used when he tries to overwrite stuff like pi, e, load_yaml(), etc...


rhaschke:
Rebased.

PR #109, opened 2015-08-12
fix #108: evaluate property blocks recursively too

### LINKED ISSUE #108: BC break: expressions and macros inside property-block definition are not expanded
```
<root version="1.0" xmlns:xacro="http://www.ros.org/wiki/xacro">
        <xacro:macro name="test">
                ahoj
        </xacro:macro>
        <xacro:property name="val" value="2" />
        <xacro:property name="prop">
                <outer attr="${val}">
                testy
                <xacro:test />
                </outer>
        </xacro:property>
        <xacro:macro name="final">
                <xacro:insert_block name="prop" />
        </xacro:macro>
        <xacro:final />
</root>
```

This snippet generates the following in Indigo:

```
<?xml version="1.0" ?>
<!-- =================================================================================== -->
<!-- |    This document was autogenerated by xacro from test.xacro
        | -->
<!-- |    EDITING THIS FILE BY HAND IS NOT RECOMMENDED
        | -->
<!-- =================================================================================== -->
<root version="1.0" xmlns:xacro="http://www.ros.org/wiki/xacro">
  <outer attr="2">

                testy


                ahoj

  </outer>
</root>
```

And the following in Jade (regardless of --inorder):

```
<?xml version="1.0" ?>
<!-- =================================================================================== -->
<!-- |    This document was autogenerated by xacro from test.xacro
        | -->
<!-- |    EDITING THIS FILE BY HAND IS NOT RECOMMENDED
        | -->
<!-- =================================================================================== -->
<root version="1.0" xmlns:xacro="http://www.ros.org/wiki/xacro">
  <outer attr="${val}">

                testy

    <xacro:test/>
  </outer>
</root>
```

rhaschke:
When switching to recursive eval_all(), I forgot to recursively evaluate the inserted block. Fixed now.
Extended the appropriate unittest to also include a property and macro definition.

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -595,7 +595,10 @@ def eval_all(node, macros, symbols):
                     raise XacroException("Undefined block \"%s\"" % name)
 
                 # cloning block allows to insert the same block multiple times
-                replace_node(node, by=block.cloneNode(deep=True), content_only=content_only)
+                block = block.cloneNode(deep=True)
+                # recursively evaluate block
+                eval_all(block, macros, symbols)
+                replace_node(node, by=block, content_only=content_only)
 
             elif is_include(node):
                 process_include(node, symbols, lambda doc: eval_all(doc, macros, symbols))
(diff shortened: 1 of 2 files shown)


peci1:
Wow, I spent a few hours figuring out where should the eval_all call land :) And I didn't succeed. Thank you very much!

PR #113, opened 2015-09-03
load_yaml now has an optional package parameter for the yaml file

thebyohazard:
Before, the yaml file loaded with load_yaml had to be in the same directory as the xacro file in which it appeared. In my use case, I have a file that needs to load a yaml file in a different package. As load_yaml was already using the ${} special command notation, I could not use a ${find package} inside of it. This might not be the best way to implement this feature, but it works and it doesn't seem to break existing stuff.

Disclaimer: I'm no python dev...

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -38,6 +38,7 @@ import re
 import sys
 import ast
 import math
+import rospkg
 
 from roslaunch import substitution_args
 from rospkg.common import ResourceNotFound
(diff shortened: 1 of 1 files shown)


thebyohazard:
I thought I had tried everything and that my desired functionality was impossible with the current code ... thus the patch. 
Based on the examples and tutorials I found on the wiki, I thought using a property necessitated using a ${} every time. So I tried 

`<xacro:property name="filename" value="$(find package)/file.yaml"/>
<xacro:property name="dictionary" value="${load_yaml(${filename})}"/>`

and similar things to no avail. I'll go ahead and answer [my question](http://answers.ros.org/question/215443/xacro-load-yaml-file-with-rospack/) with your suggestion.
Thanks

Issue #107, opened 2015-08-06
Feature: Second interface so that functionality can be used in own script


NikolasE:
I'd so far seen no way to run xacro without using the main()-function that reads sys.argv. I'd however like to convert a urdf.xacro to an urdf within my own python script. 

I am missing this functionality when I create the urdf model during the startup-process of my robots. 
I store some xacro-arguments in a database so that all robots have the same code-basis but the correct urdf is created. As the xacro/__init__py.main() assumes to be called as it's own program (using sys.argv) it's hard to include the urdf-creation can be included in my own process while I still have access to all Exceptions (e.g. if an argument was not set). 

The interface would look something like

``` python
# urdf_xacro_filename: absolute path to robot.urdf.xacro
# mappings: a dict of xacro-arguments
def convert_urdf(urdf_xacro_filename, mappings):
     (....)
     return doc.toprettyxml(indent='  ')
```

Exceptions should be re-raised so that the user of this function could handle them (e.g. just output on rospy.logginfo()). 

What do you think about such in interface? I'd be happy to implement it, if there's a good chance that it would be integrated.

rhaschke:
The functionality you request is available from a very few function calls:

``` python
opts = xacro.process_args(arg_string)
doc = xacro.parse(filename)
xacro.process_doc(doc, **vars(opts))
return doc.toprettyxml(indent='  ')
```

The essential functionality provided in xacro.main() is to handle the exceptions nicely.
As you write, you want to handle exceptions in your fashion. Hence you should kind of reimplement xacro.main() with your own exception handling.

When refactoring xacro for jade, I already tried to have as few high-level functions as possible and the above set of functions is the result. As all calls except `process_doc()` might be optional in various settings, I don't see the need for another function.

codebot:
@NikolasE  , does the functionality described by @rhaschke meet your needs?

PR #122, opened 2015-12-03
Added a more useful Python API.

peci1:
You can now easily use xacro from within Python the same way you would use it from the CLI.


rhaschke:
This is related to #107. I still don't see the need for an API change. The [existing set of functions](https://github.com/ros/xacro/issues/107#issuecomment-128503487)
allows to flexibly process xacro from within python. 
@peci1 Where do you see the benefits of your convert function?

Edit: As a downside of your proposed method, the embedding python code has no means for own exception handling and it would need to parse the generated string again for further processing if needed.

peci1:
The reason is the following: to use xacro from bash, you just call one script. Let's make it that simple in Python, too. Not all users want to dive into the inner workings of xacro and extract the needed lines of code themselves. Because of the use of globals, they can't even be sure they called everything needed until they go through the code very thoroughly.

Having this simplified Python API is IMO no BC break, no additional burden to the code, just  separating two unrelated things (the procedure needed to process an input file to an output vs. handling errors on the commandline). 

An advantage of the proposed API is then a more readable way of passing the `process_doc` parameters. The current code parses them from an argv-like string, which really smells (at least for me).

Using your 4 lines of code from #107 is of course easy, but I can't see a reason for not having them wrapped up in one function and simplified so that the argv-like line parsing is not needed.

rhaschke:
If your only concern is the option parsing, you can simplify the code even more (`process_doc` already provides all desired options as proper arguments):

``` python
doc = xacro.parse(filename)
xacro.process_doc(doc)
return doc.toprettyxml(indent='  ')
```

I agree, that these lines should be documented somewhere. Where would you look for that information?
The benefit of your proposal is that the exception handling is integrated in `convert()`. I will propose a different solution then.

PR #125, opened 2015-12-04
refactored main(), fix #122, #107

rhaschke:
This is an alternative implementation for #122, also resolving #107.

Having two similar requests, I finally agreed to refactor main() and move the main processing pipeline to a separate function `process_file(filename, **kwargs)` facilitating reuse from other python code.
In order to have not only three function calls there, I decided also to move the `auto-generated` banner to this function. This increases readability of `main()` too. However, using the `--includes` option, this will also create the banner in the output, which changes previous behaviour. I guess, that's ok.
@peci1 [agreed to that approach too](https://github.com/tradr-project/xacro/pull/1#issuecomment-161984056).


rethink-imcmahon:
+1 Looks good to me. Refactoring for easy use as a python library seems like a worthy goal. What do you mean by

> using the --includes option, this will also create the banner in the output, which changes previous behaviour

What behavior would this change exactly?

rhaschke:
The command-line option `--includes` was used to process only all `<include>` directives and return the otherwise unprocessed document. I guess this was for debugging.
So far, this pre-processed document was output without the xacro banner, which will change with this PR.

PR #126, opened 2015-12-09
Unicode processing

codebot:
This PR attempts to permit at least some level of Unicode characters in xacro. Since xacro property and macro names eventually become Python identifiers, those strings are restricted to ASCII (like Python identifiers), but the rest of the content should now allow Unicode, as was requested in #69 . Feedback appreciated though; it's quite possible I missed some code paths, and I'm sure there is a better way to do some of these things.

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -47,10 +47,11 @@ from .xmlutils import *
 from .cli import process_args
 
 
-try:
+try: # python 2
     _basestr = basestring
-except NameError:
-    _basestr = unicode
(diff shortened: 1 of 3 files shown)


rhaschke:
Hi Morgan, what about this PR? Is there any reason why it isn't yet merged?
Meanwhile it is out-of-date and doesn't merge into jade-devel anymore. However, there is an easy fix.
Should I rebase and open a new PR?

rhaschke:
Rebased and created new PR #139.

PR #138, opened 2016-07-12
Fix file check for python3.

hgaiser:
Running xacro on arch with Python 3.5 I got the following error :

NameError: name 'file' is not defined

'file' is no longer defined in python3, comparing it to io.IOBase should work according to http://stackoverflow.com/questions/18880948/python-3-3-2-check-that-object-is-of-type-file

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -902,7 +902,7 @@ def parse(inp, filename=None):
     try:
         if isinstance(inp, _basestr):
             return xml.dom.minidom.parseString(inp)
-        elif isinstance(inp, file):
+        elif hasattr(inp, 'read'):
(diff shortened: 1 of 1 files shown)


hgaiser:
Changed the check to a python2/3 independent check.

http://stackoverflow.com/questions/3783995/how-do-you-check-an-object-is-an-instance-of-file

rhaschke:
Syntax error! You missed a colon.
Could you please squash everything into a single commit?

Otherwise +1.

hgaiser:
Done and done :).

rhaschke:
Thanks. Checks pass as well now ;-)
Morgan, from my point of view, this could be merged.

PR #140, opened 2016-07-21
Add debug option.

130s:
Undefined property error is hard to debug without knowing which properties are read / are not read. So I'm adding debug feature.

```
rosrun xacro xacro foo.xacro --debug

2016-07-20 17:57:34,024 - DEBUG - Input xacro file full path=/home/roslinguini/src/sirobot1_gazebo/model/foo.xacro
2016-07-20 17:57:34,119 - DEBUG - name=pi, val=3.1415926535897931
2016-07-20 17:57:34,119 - DEBUG - name=small_mass, val=0.001
2016-07-20 17:57:34,120 - DEBUG - name=pi, val=3.1415926535897931
2016-07-20 17:57:34,120 - DEBUG - name=pi, val=3.1415926535897931
2016-07-20 17:57:34,120 - DEBUG - name=width_shelf, val=0.005
2016-07-20 17:57:34,120 - DEBUG - name=radius, val=0.002
2016-07-20 17:57:34,120 - DEBUG - name=pi, val=3.1415926535897931
2016-07-20 17:57:34,120 - DEBUG - name=pi, val=3.1415926535897931
2016-07-20 17:57:34,121 - DEBUG - name=wheel_mass, val=0.5
2016-07-20 17:57:34,121 - DEBUG - name=sphere_mass, val=0.3
```

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -34,6 +34,7 @@ from __future__ import print_function
 
 import getopt
 import glob
+import logging
 import os
 import re
 import string
@@ -342,6 +343,7 @@ def grab_properties(doc):
                                  '"{", "}", or "$" : "' + name + '"')
             else:
                 table[name] = value
+                logging.debug('name={}, val={}'.format(name, table[name]))
 
             elt.parentNode.removeChild(elt)
             elt = None
@@ -645,11 +647,12 @@ def open_output(output_filename):
 
 def main():
     try:
-        opts, args = getopt.gnu_getopt(sys.argv[1:], "ho:", ['deps', 'includes'])
+        opts, args = getopt.gnu_getopt(sys.argv[1:], "ho:", ['debug', 'deps', 'includes'])
     except getopt.GetoptError as err:
         print(str(err))
         print_usage(2)
 
+    debug_mode = False
     just_deps = False
     just_includes = False
 
@@ -659,6 +662,8 @@ def main():
             print_usage(0)
         elif o == '-o':
             output_filename = a
+        elif o == '--debug':
+            debug_mode = True
         elif o == '--deps':
             just_deps = True
         elif o == '--includes':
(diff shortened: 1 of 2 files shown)


rhaschke:
You should consider to switch to the new xacro version of >= Jade, which is also backported to Indigo and available via new option --inorder. This new version has extensive debugging features with options `-v` ... `-vvv`. IMHO the Indigo branch is deprecated.

130s:
tldr; Closing since a workaround is provided that should work great (only offline for Indigo >= though).

> You should consider to switch to the new xacro version of >= Jade, which is also backported to Indigo and available via new option --inorder. 

I guess you're suggesting to check offline the xacro file using xacro from >= Jade distro. I see the point after looking at the wiki http://wiki.ros.org/xacro#Processing_Order

`inorder` isn't available during runtime on Indigo => distro. I guess this feature will never be available there?

> IMHO the Indigo branch is deprecated.

I don't know if that is acceptable as an end-user's perspective, since ROS Indigo itself is supported until 2019...

rhaschke:
> > IMHO the Indigo branch is deprecated.
> 
> I don't know if that is acceptable as a end-user since ROS Indigo itself is supported until 2019...

Of course ROS Indigo will be further supported and bug fixes will be performed.
However, you are requesting a new feature for Indigo ;-)

130s:
> > > ```
> > >    IMHO the Indigo branch is deprecated.
> > > ```
> > 
> > ```
> > I don't know if that is acceptable as a end-user since ROS Indigo itself is supported until 2019...
> > ```
> 
> Of course ROS Indigo will be further supported and bug fixes will be performed.
> However, you are requesting a new feature for Indigo ;-)

I believe that it is NOT commonly accepted idea that new feature won't be added to older distribution (e.g. Indigo).
As long as a distro is supported and new feature development is happening on its repository, I think it's natural for many users to expect that they'll keep getting those features.

I'm not forcing xacro maintainers to backport new features into Indigo, but I just wanted to point that out.

PR #146, opened 2017-02-19
Convert exception to string in a python2/3 compatible way.

de-vri-es:
In python3, `Exception` doesn't have a `message` attribute anymore. The proper way to convert exceptions to strings is just to call `str` on them (works for both python 2 and 3). We can't do that directly here, because we want to to use the string conversion method of the parent class. We can still do that with `super()` though. So that is what this PR does.

Without it, the exception handler crashes without printing the wanted message.

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -129,8 +129,8 @@ class XacroException(Exception):
         self.macros = [] if macro is None else [macro]
 
     def __str__(self):
-        return ' '.join([str(item) for item in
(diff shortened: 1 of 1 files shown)


rhaschke (approved):
LGTM.

de-vri-es:
Any blocking issues? I need to keep a patched version locally now to get meaningful error messages from xacro :(

de-vri-es:
Awesome, thanks @codebot and @rhaschke

Issue #145, opened 2017-01-30
[xacro indigo] True/False vs. true/false when using conditional blocks with --inorder


fmessmer:
When testing the new possibilities that come with using the `--inorder` argument with xacro under indigo, we found the following behaviour - **this possibly is a regression:**
When using xacro:properties or macro parameters within a conditional expression e.g using `or`, the result is not as expected when the values of the properties are set to `true`/`false`. The result is correct when using `True`/`False`, though.

In [xacro](http://wiki.ros.org/action/fullsearch/xacro#Conditional_Blocks) the documentation says:
 > The expression needs to evaluate to "0", "1", "true", or "false", otherwise an error will be thrown.

However, no error is thrown...the block just seems to be skipped.
Documentation also says:
 > The more powerful evaluation capabilities in ROS Jade allow for much more complex expression. Virtually any python expression that evaluates to a Boolean is feasible:

Which is consistent with our observations as in python it's not `true`/`false` but `True`/`False`....but this is inconsistent with what is said in the first quote.

A little example:
When running `rosrun xacro xacro --inorder` on the following snippet:
```
<?xml version="1.0"?>
<robot xmlns:xacro="http://www.ros.org/wiki/xacro">

  <xacro:property name="cond1" value="false"/>
  <xacro:property name="cond2" value="true"/>

  <xacro:if value="${cond1 or cond2}">
    <link name="test"/>
  </xacro:if>

</robot>
```
The resulting URDF looks like this:
```
<?xml version="1.0" ?>
<!-- =================================================================================== -->
<!-- |    This document was autogenerated by xacro from conditional_block.urdf.xacro   | -->
<!-- |    EDITING THIS FILE BY HAND IS NOT RECOMMENDED                                 | -->
<!-- =================================================================================== -->
<robot xmlns:xacro="http://www.ros.org/wiki/xacro">
</robot>
```

While running the same command on the following snippet:
```
<?xml version="1.0"?>
<robot xmlns:xacro="http://www.ros.org/wiki/xacro">

  <xacro:property name="cond1" value="False"/>
  <xacro:property name="cond2" value="True"/>

  <xacro:if value="${cond1 or cond2}">
    <link name="test"/>
  </xacro:if>

</robot>
```
produces the expected result:
```
<?xml version="1.0" ?>
<!-- =================================================================================== -->
<!-- |    This document was autogenerated by xacro from conditional_block.urdf.xacro   | -->
[...]

PR #150, opened 2017-04-24
True/False literals in expressions

rhaschke:
As indicated in #145, python's True / False literals are not correctly forwarded to xacro, which doesn't allow for the following xacro code:
```
<robot xmlns:xacro="http://www.ros.org/wiki/xacro">
  <xacro:property name="cond" value="${True}"/>
  <xacro:if value="${cond}"><enabled/></xacro:if>
</robot>
```
This PR adds True/False to the list of available python keywords in `${}` expressions.


rhaschke:
@codebot I think, this patch should be ported to lunar-devel and back-ported up to indigo?

jliviero:
@codebot / @rhaschke, can you clarify whether this fix was actually ported to indigo/jade/lunar?  The issue #145 was closed, but it references indigo specifically.

PR #158, opened 2017-08-29
Small fixes to make it work with Python 3

kartikmohta:
Note that two tests fail due to internal changes in Python 3:
- test_iterable_literals_eval: The order of keys in dict changed in Py3
- test_no_evaluation: ast.literal_eval('5 -2') gives 3 in Py3

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -950,7 +950,7 @@ def process_doc(doc,
     if do_check_order and symbols.redefined:
         warning("Document is incompatible to --inorder processing.")
         warning("The following properties were redefined after usage:")
-        for k, v in symbols.redefined.iteritems():
+        for k, v in symbols.redefined.items():
             message(k, "redefined in", v, color='yellow')
 
 
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -203,17 +203,13 @@ class Table(object):
     @staticmethod
     def _eval_literal(value):
         if isinstance(value, _basestr):
-            try:
-                # try to evaluate as literal, e.g. number, boolean, etc.
-                # this is needed to handle numbers in property definitions as numbers, not strings
-                evaluated = ast.literal_eval(value.strip())
-                # However, (simple) list, tuple, dict expressions will be evaluated as such too,
-                # which would break expected behaviour. Thus we only accept the evaluation otherwise.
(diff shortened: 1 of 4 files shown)


rhaschke:
~~I'm wondering whether these changes are really neccessary. Did you have specific issues running xacro with python3? Running the unittests on my side, I don't see any issues.~~
Sorry, the unittests didn't yet used python3 on my side. Now, I'm getting some errors too ;-)

kartikmohta:
`cStringIO` and `dict.iteritems` are not present in Python 3, so these changes are required.

src/xacro/__init__.py
```diff
@@ -950,7 +950,7 @@ def process_doc(doc,
     if do_check_order and symbols.redefined:
         warning("Document is incompatible to --inorder processing.")
         warning("The following properties were redefined after usage:")
-        for k, v in symbols.redefined.iteritems():
+        for k, v in symbols.redefined.items():
```
rhaschke, line 949: Although iteritems() is more efficient than items() in python2, I think for the sake of simplicity, we can switch to items().

xacro.py
```diff
@@ -50,7 +50,7 @@
 this_dir_cwd = os.getcwd()
 os.chdir(cur_dir)
 # Remove this dir from path
-sys.path = filter(lambda a: a not in [this_dir, this_dir_cwd], sys.path)
+sys.path = [a for a in sys.path if a not in [this_dir, this_dir_cwd]]
```
rhaschke, line 53: This is definitely more pythonic ;-)

rhaschke (changes requested):
Additionally to your requested changes, I had to
import xml.dom.minidom
in xmlutils.py. Strange, that you didn't notice that issue.

src/xacro/xmlutils.py
```diff
@@ -31,6 +31,7 @@
 # Maintainer: Morgan Quigley <morgan@osrfoundation.org>
 
 import xml
```
rhaschke: The "import xml" can be removed then.

rhaschke:
As soon as #161 is merged, I will add some more commits, cleanup and re-issue as a new PR.
I fixed everything for python3 already, such that the unittests pass:
https://travis-ci.org/ubi-agni/xacro

PR #160, opened 2017-09-13
Travis trusty

codebot:
Updating the Travis config to explicitly request Trusty and use Indigo as the base ROS environment.


rhaschke:
@codebot I guess, you want to have this PR w.r.t. to lunar-devel branch instead of kinetic-devel first.
Otherwise you pull in the version bump.

PR #163, opened 2017-09-15
python 3 compatibility

rhaschke:
This is a replacement for #158, fixing (most) remaining issues for python 3.

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -950,7 +950,7 @@ def process_doc(doc,
     if do_check_order and symbols.redefined:
         warning("Document is incompatible to --inorder processing.")
         warning("The following properties were redefined after usage:")
-        for k, v in symbols.redefined.iteritems():
+        for k, v in symbols.redefined.items():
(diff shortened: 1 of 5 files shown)


src/xacro/__init__.py
```diff
@@ -577,6 +573,9 @@ def grab_properties(elt, table):
         elt = next
 
 
+# TODO: The order of arguments here is relevant!
```
rhaschke: I really like the beauty and simplicity of this lexer definitions.
However it relies on ordered keyword arguments, which are not guaranteed for python 3.0 - 3.5.
Nevertheless, personally I would blame python to be buggy here (having changed the semantics).
As this is fixed in python 3.6, I wouldn't go for an ugly workaround.

The issue only arises for documents containing `$${` expressions. We could also solve the issue by finding regular expressions that are mutually exclusive.

rhaschke: Actually, it wasn't that difficult to come up with new regular expressions.
Next commit also fixes some corner-case issues that were failing before ;-)

PR #167, opened 2017-10-05
allow to specify extra dependencies for xacro_add_xacro_file(s)

rhaschke:
Allowing to include variable-generated files and supporting load_yaml(),
introduces dependencies that cannot be catched by xacro --deps.
The new argument DEPENDS allows to specify any other files required.

Sleeping a night over this issue, it came clear to me that it would be much better to directly list files loaded with load_yaml in the dependencies too. That's what the second commit achieves. However, I think manually added dependencies should be still possible in case...


rhaschke:
I will reopen this PR using a new branch name...

PR #173, opened 2017-11-16
allow default values for properties

rhaschke:
This PR allows for definition of default values for properties (much like it is possible for arguments):

```xml
<a xmlns:xacro="http://www.ros.org/xacro">
	<xacro:property name="prop" default="false"/>
	<xacro:unless value="${prop}">
		<foo/>
		<xacro:property name="prop" value="true"/>
	</xacro:unless>

	<!-- second foo should be ignored -->
	<xacro:unless value="${prop}">
		<foo/>
		<xacro:property name="prop" value="true"/>
	</xacro:unless>
</a>
```

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -531,9 +531,20 @@ def grab_property(elt, table):
     assert(elt.tagName in ['property', 'xacro:property'])
     remove_previous_comments(elt)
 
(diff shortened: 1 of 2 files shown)


rhaschke:
@guihomework I think this solves your issue ;-)

PR #174, opened 2017-11-16
Merge changes from kinetic-devel to lunar-devel branch

kartikmohta:
The fixes would be good to have in lunar as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,9 @@
 dist: trusty
-sudo: required
+sudo: enabled
 language: python
 python:
-  - "2.7_with_system_site_packages"
+  - "2.7"
+
 # command to install dependencies
 install:
   - pip install nose coverage pep8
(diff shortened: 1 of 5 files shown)


rhaschke:
@codebot Of course we should have those fixes in Lunar as well.
@kartikmohta However, instead of merging, I would suggest some cherry-picks. Merging will also merge version-number changes. Could you, @kartikmohta, please change your PR to use cherry-picks instead?

kartikmohta:
@rhaschke Looking at the file diffs, it doesn't look like there are any version changes. I think a squash and merge would be fine since the diff is small.
BTW, is there any specific need to keep kinetic and lunar as different branches?

rhaschke:
You are right. Merging will not affect any version numbers in package.xml. So, I don't mind to merge as well. But, it's up to @codebot to decide about the merging/rebasing strategy ;-)
Regarding the different branches: Yes, we need different branches as kinetic and lunar have different version numbers. (However, the merging commit doesn't change them).

rhaschke:
I decided to go for cherry picks.

PR #186, opened 2018-05-02
fix parsing of parameter defaults surrounded with single exclamation …

raphaeldeimel:
This patch fixes parsing of parameter defaults that are surrounded by single exclamation marks (e.g. to escape space characters)

Example: params="xyz:='0 0 0'" 

Without the patch, parameter values keep the exclamation marks.
With the patch, parsing behaves like in the lunar release (exclamation marks are stripped)

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -481,6 +481,9 @@ def parse_macro_arg(s):
         # there is a default value specified for param
         param, forward, default, rest = m.groups()
         if not default: default = None
+        if default.startswith("'") and default.endswith("'"):
+            default = default[1:-1]
+        print(param, default)
         return param, (param if forward else None, default), rest
     else:
         # there is no default specified at all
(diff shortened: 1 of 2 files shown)


rhaschke:
Thanks for pointing out this issue. Could you please also provide a unittest in [`test/test_xacro.py, class TestXacro`](https://github.com/ros/xacro/blob/lunar-devel/test/test_xacro.py#L296)?
I'm not yet sure, your solution is the right approach. Are you running xacro on python 2 or 3?

raphaeldeimel:
I was using xacro on python2.

Frankly, I  think my patch is a bit of a hack, the exclamation marks shouldn't be in the match of the parsing regex in the first place. But with the way the regex is formulated right now it is not trivial to ignore them, which in turn is probably a result of using a regex to parse context-free grammar construct...

rhaschke:
Closing this in favour of #187

PR #194, opened 2018-10-19
Ros2 devel

swatifulzele:
hello


rhaschke:
Thank you very much for this contribution. However, I'm not sure yet how to deal with this. I'm not very familar with ROS2 yet. But, obviously there are incompatible changes in the make system (catkin -> ament) as well as an incompatible package name change (roslaunch -> roslauch2).

rhaschke:
Dear @codebot @wjwwood,
the common way to handle ROS1 -> ROS2 transition is to clone the repo on github.com/ros2/xacro.
Could anybody with appropriate rights please do so and grant me maintainer status there as well?

The main issue for migrating to ROS2 seems to be the build system (`catkin` -> `ament`). However, I have read about [`catment`](https://github.com/ros2/ros2/wiki/catment) as well. According to this Wiki, it should be possible to stick with the catkin-based cmake file and package.xml in this package, and still successfully build with ament. What's the status of this catment thing? Is it well established practice or alpha testing only?

Is there a way to specify different package dependencies for ROS1 and ROS2 in `package.xml` (as required here for `ros2launch` opposed to `roslaunch`)?

PR #203, opened 2018-12-19
Ros2

vandanamandlik:
Ported xacro package to ROS2 and tested it on ROS2 version bouncy and crystal.
@rhaschke, Could you please review it and give your feedback ?


darkdragon-001:
Is there any chance this is reviewed and released soon?

rhaschke:
Sorry for being unresponsive on this. I am too busy with other projects and I didn't familiarize myself with ROS2 yet. Hopefully, I can find some time during my summer holidays, i.e. during the next 4 weeks to work on this.
To facilitate the review and merge process, I suggest to split this large PR into individual commits as follows:
- ROS-version agnostic, i.e. general improvements of xacro
  I already have put some effort into making xacro python2/3 agnostic as well as to use it without ROS at all. These properties should be maintained.
  - cosmetics like flake, pep
- ROS2-specific changes
  - adaptions for the ROS2 build
    As far as I understand, [colcon can build catkin packages as well](https://colcon.readthedocs.io/en/released/user/quick-start.html#build-ros-1-packages). So, do we really need such adaptions?
  - replacement for ROS1 substitution args `$(...)`

Would be great, if you, @vandanamandlik or @kazuki0824, could prepare this.

schornakj:
Is there an example somewhere showing how to parse a URDF xacro file in ROS 2, for example from within a .launch.py script?

mogumbo:
I would also like to see an example of using a xacro file in a launch.py script. I have been trying to use it in dashing like this:
 xacro_name = os.path.join(get_package_share_directory('my_package'), 'urdf/rp1a_robot.xacro')
 urdf = xacro.process_file(xacro_name)

This errors on a call to get_path() in substitution_args.py. I can get past that by importing rospkg and uncommenting some code in _get_rospack(). But then I'm just stuck on another error.

mogumbo:
I would also like to see an example of using a xacro file in a launch.py script. I have been trying to use it in dashing like this:
xacro_name = os.path.join(get_package_share_directory('my_package'), 'urdf/my_robot.xacro')
urdf = xacro.process_file(xacro_name)

This errors on a call to get_path() in substitution_args.py. I can get past that by importing rospkg and uncommenting some code in _get_rospack(). But then I'm just stuck on another error.

schornakj:
> I would also like to see an example of using a xacro file in a launch.py script. I have been trying to use it in dashing like this:
> xacro_name = os.path.join(get_package_share_directory('my_package'), 'urdf/my_robot.xacro')
> urdf = xacro.process_file(xacro_name)
> 
> This errors on a call to get_path() in substitution_args.py. I can get past that by importing rospkg and uncommenting some code in _get_rospack(). But then I'm just stuck on another error.

Your method should work, but the underlying issue (as @rhaschke mentioned) is that the version of `substitution_args.py` included in this PR isn't really ROS2 compatible. Most robot `.urdf.xacro` files heavily rely on substitution to find things like collision meshes in support packages, and since `substitution_args.py` is still using Catkin methods in some places it makes sense that it wouldn't work in a ROS2 system.

gonzodepedro:
Changes on #211

Issue #205, opened 2019-03-01
catkin_python_setup called twice


kejxu:
in [this change](https://github.com/ros/xacro/commit/00f05341dce0e6330916282bc9176c81795fdf74#diff-af3b638bc2a3e6c650974192a53c7291), a second `catkin_python_setup` is added.

Is this intended? If so, just curious... what is the justification behind the change?

rhaschke:
Fixed via #206.

PR #206, opened 2019-03-01
remove duplicate catkin_python_setup()

kejxu:
this change aims to address https://github.com/ros/xacro/issues/205, by reducing the number of `catkin_python_setup()` calls to only 1


rhaschke (approved):

rhaschke:
Thanks for reporting and fixing this.

PR #208, opened 2019-03-04
Add `len()` to allowed Python builtins

peci1:
Allows to create a loop like this one:

    <xacro:macro name="include_payload_xacros" params="xacros">
        <xacro:if value="${len(xacros) > 0}">
            <xacro:include filename="${xacros[0]}" />
            <xacro:include_payload_xacros xacros="${xacros[1:]}" />
        </xacro:if>
    </xacro:macro>
    <xacro:include_payload_xacros xacros="${$(arg payload_xacros)}" />

Without this PR, you need to use the `list.__len__()` function, which is not so nice, but it also works.


peci1:
That was lightning fast, thanks :)

rhaschke:
You are welcome. Simple things are fast to merge :wink:

PR #221, opened 2019-10-07
Dashing pep fixes

gonzodepedro:
The idea is to bring some changes I made into this branch

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -23,7 +23,6 @@
   <depend>ament_index_python</depend>
   <depend>python3-yaml</depend>
   <test_depend>ament_lint_auto</test_depend>
-  <test_depend>ament_cmake_pytest</test_depend>
 
   <export>
     <build_type>ament_cmake</build_type>
--- a/xacro/__init__.py
+++ b/xacro/__init__.py
@@ -142,7 +142,6 @@ class Macro(object):
         self.defaultmap = {}  # default parameter values
         self.history = []  # definition history
 
-
 def eval_extension(s):
     if s == '$(cwd)':
         return os.getcwd()
@@ -599,7 +598,7 @@ def handle_macro_call(node, macros, symbols):
     for name, value in node.attributes.items():
         if name not in params:
             raise XacroException(
-                'Invalid parameter "%s"' % unicode(name), macro=m)
+                "Invalid parameter '%s'" % unicode(name), macro=m)
         params.remove(name)
         scoped._setitem(name, eval_text(value, symbols), unevaluated=False)
         node.setAttribute(name, "")  # suppress second evaluation in eval_all()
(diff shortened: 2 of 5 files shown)


package.xml
```diff
@@ -23,7 +23,6 @@
   <depend>ament_index_python</depend>
   <depend>python3-yaml</depend>
   <test_depend>ament_lint_auto</test_depend>
-  <test_depend>ament_cmake_pytest</test_depend>
```
rhaschke, line 26: Why did you remove this dependency?

xacro/__init__.py
```diff
@@ -48,7 +48,6 @@
 # Dictionary of substitution args
 substitution_args_context = {}
 
-
```
rhaschke, line 51: Doesn't PEP8 require these two line breaks?

xacro/__init__.py
```diff
@@ -143,7 +142,6 @@ def __init__(self):
         self.defaultmap = {}  # default parameter values
         self.history = []  # definition history
 
-
```
rhaschke, line 146: Same here.

xacro/__init__.py
```diff
@@ -298,7 +296,7 @@ def next(self):
 
 
 all_includes = []
-include_no_matches_msg = """Include tag's filename spec \"{}\" matched no files."""
+include_no_matches_msg = 'Include tag\'s filename spec "{}" matched no files.'
```
rhaschke, line 299: I'm not sure why do you want to change most of the quotes. I don't see a benefit in these changes if they don't simplify anything. Here, for example, you just move the backslash from the double to the single quote(s).

rhaschke (commented):
Thanks for continuing your work on xacro. I extracted all acceptable changes in #222 and merged them.
For all remaining changes (mostly changing quotes), I would like to get a better motivation from you.
Notice, that by accident you introduced a breaking change (using `unicode` instead of `str`).

Please, in future, avoid pushing PR branches to the main repo directly, but create your own fork.
All branches created on the main repo, introduce additional noise on Travis CI.

xacro/__init__.py
clalancette: The only reason to change most of the quotes is that one of the default ament python linters insists on single-quotes for most everything.  If you don't think that is worthwhile, then we won't be able to enable the linters.  If that's the case, that's fine, we can just drop most of this PR (though I still think it would be nice to remove the Python 2 fallback code).  Let us know what you want to do here.

rhaschke: I already merged the uncritical parts of this PR, particularly the removal of the Python 2 fallback code. Regarding quotes, I'm hesitating to change all the code and risk new issues. If the ament linter is too restrictive, we shouldn't use it. Apart from the quotes there are some other pep8 issues (too long lines, because the xacro code is formated for 127 chars line width, while the ament linter checks for 99 chars only), which I don't want to resolve.

clalancette: All right, that's totally fine with me.  In that case, I think we can close this out.

PR #229, opened 2019-11-25
Enable launching xacro easily from ros2 launch files

gonzodepedro:
This PR creates a 'process' method in xacro to enable launching it from launch files, like:

`xacro.process(urdf_file, robot_urdf)`

### DIFF
--- a/xacro/__init__.py
+++ b/xacro/__init__.py
@@ -1034,11 +1034,6 @@ def _process(input_file_name, opts):
         out.close()
 
 
-def main():
-    opts, input_file_name = process_args(sys.argv[1:])
-    _process(input_file_name, vars(opts))
-
-
 def process(input_file_name, just_deps=False, xacro_ns=True, verbosity=1, mappings={}):
     """Function to be used from python code, returning the processed XML"""
     from io import StringIO
@@ -1048,3 +1043,8 @@ def process(input_file_name, just_deps=False, xacro_ns=True, verbosity=1, mappin
     result = sys.stdout.read()
     sys.stdout = old  # restore sys.stdout
     return result
+
+
+def main():
+    opts, input_file_name = process_args(sys.argv[1:])
+    _process(input_file_name, vars(opts))
--- a/xacro/__init__.py
+++ b/xacro/__init__.py
@@ -1044,7 +1044,6 @@ def exec(input_file_name, opts):
             sys.exit(2)  # gracefully exit with error condition
 
     # special output mode
-    #if opts.just_deps:
     if opts['just_deps']:
         out.write(' '.join(set(all_includes)))
         out.write('\n')
--- a/xacro/__init__.py
+++ b/xacro/__init__.py
@@ -991,11 +991,11 @@ def process_file(input_file_name, **kwargs):
     return doc
 
 
-def process(input_file_name, output=None, in_order=True, do_check_order=False,
+def process(input_file_name, output_file_name, in_order=True, do_check_order=False,
             just_deps=False, just_includes=False, xacro_ns=True, verbosity=1, mappings={}):
 
     opts_map = {
-        'output': output,
+        'output': output_file_name,
         'in_order': in_order,
         'do_check_order': do_check_order,
         'just_deps': just_deps,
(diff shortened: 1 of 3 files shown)


xacro/__init__.py
```diff
@@ -1026,14 +1047,16 @@ def main():
             sys.exit(2)  # gracefully exit with error condition
 
     # special output mode
-    if opts.just_deps:
-        out.write(" ".join(set(all_includes)))
+    #if opts.just_deps:
+    if opts['just_deps']:
+        out.write(' '.join(set(all_includes)))
         print()
```
rhaschke: What happens to those prints on stdout?

xacro/__init__.py
```diff
@@ -991,24 +991,45 @@ def process_file(input_file_name, **kwargs):
...
+        'just_deps': just_deps,
+        'just_includes': just_includes,
+        'xacro_ns': xacro_ns,
+        'verbosity': verbosity,
+        'mappings': mappings,
+    }
+
+    return exec(input_file_name, opts_map)
+
 def main():
```
rhaschke: Can you move main() to the bottom of the file, please?

xacro/__init__.py
```diff
@@ -991,24 +991,45 @@ def process_file(input_file_name, **kwargs):
...
+    except ExpatError as e:
+        error('XML parsing error: %s' % str(e), alt_text=None)
         if verbosity > 0:
             print_location(filestack, e)
             print(file=sys.stderr)  # add empty separator line before error
             print('Check that:', file=sys.stderr)
             print(' - Your XML is well-formed', file=sys.stderr)
             print(' - You have the xacro xmlns declaration:',
-                  'xmlns:xacro="http://www.ros.org/wiki/xacro"', file=sys.stderr)
+                  'xmlns:xacro=\"http://www.ros.org/wiki/xacro\"', file=sys.stderr)
```
rhaschke: Why do you need to escape double quotes here? This makes it (slightly) harder to read the string.

xacro/__init__.py
```diff
@@ -1026,14 +1047,16 @@ def main():
...
 
     # special output mode
-    if opts.just_deps:
-        out.write(" ".join(set(all_includes)))
+    #if opts.just_deps:
+    if opts['just_deps']:
+        out.write(' '.join(set(all_includes)))
         print()
-        return
+        return ' '.join(set(all_includes))
```
rhaschke: Why do you want to create the output twice (here as a return value) and in line 1052 as output to `out`.
Same applies for the way more costly `doc.toprettyxml` below.
I think, as a user you should decide whether to use the python function interface and use the return value or call the xacro script and use the output generated.

rhaschke (changes requested):
I don't see the use case yet, but in general I'm fine with this. Please address the comments.

xacro/__init__.py
gonzodepedro: I could think of a use case to use the output from within a python launch script. But that no longer makes sense. I think that the best thing is to force the user to write output to a file when called from a script instead of returning the output.

xacro/__init__.py
```diff
@@ -1026,14 +1044,22 @@ def main():
             sys.exit(2)  # gracefully exit with error condition
 
     # special output mode
-    if opts.just_deps:
-        out.write(" ".join(set(all_includes)))
-        print()
+    #if opts.just_deps:
```
clalancette: Nit: remove this commented-out line.

clalancette (commented):
I have one nit to fix, but this looks generally like an improvement to me for launching.

xacro/__init__.py
gonzodepedro: Thanks for your review

xacro/__init__.py
rhaschke: If you are forcing the user to write an output file, he can use xacro the old way, i.e. launch a sub process, as well. I thought, the main goal of this PR was to enable direct calling of `xacro.process()` and to utilize its _return value_ in the launch script. So, I expected that you would have removed the file output, but not the return value. 
I have implemented this idea in the last amended commit.
Please clarify your intent.

rhaschke:
@gonzodepedro I have added some cleanup commits. Most importantly, you re-introduced some obsolete options. Please also clarify your intend for this PR (see https://github.com/ros/xacro/pull/229#discussion_r355176383).

xacro/__init__.py
gonzodepedro: " direct calling of xacro.process() and to utilize its return value in the launch script" sounds like the way to go! Thanks

PR #232, opened 2020-03-23
handling non optional existence in xacro files

veimox:
**Use case:**
Being able to load files is currently supported either in xacro files, but these files are required to exist otherwise an exception is thrown. I want to be able to include those files if they exist but if they dont, just continue. This would allow me to override properties from a hardcoded path only if needed like (e.g. ~/.ros/my-custom-calibration.xacro)

**Comments**
I have added the _optional_ support to xacro files and added the unit testing for them. Besides i am merging this in `kinetic-devel` only, we can port it later to `melodic-devel` and friends.

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -362,12 +362,14 @@ def is_include(elt):
     return True
 
 
-def get_include_files(filename_spec, symbols):
+def get_include_files(filename_spec, symbols, optional=False):
     try:
         filename_spec = abs_filename_spec(eval_text(filename_spec, symbols))
     except XacroException as e:
         if e.exc and isinstance(e.exc, NameError) and symbols is None:
             raise XacroException('variable filename is supported with --inorder option only')
+        elif e.exc.strerror == "No such file or directory" and optional is True:
+            return
         else:
             raise
 
@@ -402,7 +404,7 @@ def import_xml_namespaces(parent, attributes):
 
 def process_include(elt, macros, symbols, func):
     included = []
-    filename_spec, namespace_spec = check_attrs(elt, ['filename'], ['ns'])
+    filename_spec, namespace_spec, optional = check_attrs(elt, ['filename'], ['ns', 'optional'])
     if namespace_spec:
         try:
             namespace_spec = eval_text(namespace_spec, symbols)
@@ -413,18 +415,27 @@ def process_include(elt, macros, symbols, func):
         except TypeError:
             raise XacroException('namespaces are supported with --inorder option only')
 
-    for filename in get_include_files(filename_spec, symbols):
+    optional = True if optional == "True" else False
+
+    for filename in get_include_files(filename_spec, symbols, optional):
         # extend filestack
         oldstack = push_file(filename)
-        include = parse(None, filename).documentElement
+        try:
+            include = parse(None, filename).documentElement
 
-        # recursive call to func
-        func(include, macros, symbols)
-        included.append(include)
-        import_xml_namespaces(elt.parentNode, include.attributes)
+            # recursive call to func
+            func(include, macros, symbols)
+            included.append(include)
+            import_xml_namespaces(elt.parentNode, include.attributes)
 
-        # restore filestack
-        restore_filestack(oldstack)
+            # restore filestack
+            restore_filestack(oldstack)
+
+        except XacroException as e:
+            if e.exc.strerror == "No such file or directory" and optional is True:
+                continue
+            else:
+                raise
 
     remove_previous_comments(elt)
     # replace the include tag with the nodes of the included file(s)
@@ -914,7 +925,7 @@ def parse(inp, filename=None):
         except IOError as e:
             # do not report currently processed file as "in file ..."
             filestack.pop()
-            raise XacroException(e.strerror + ": " + e.filename)
+            raise XacroException(e.strerror + ": " + e.filename, exc=e)
 
     try:
         if isinstance(inp, _basestr):
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -415,25 +415,25 @@ def process_include(elt, macros, symbols, func):
 
     optional = True if optional == "True" else False
 
-    try:
-        for filename in get_include_files(filename_spec, symbols):
-            # extend filestack
-            oldstack = push_file(filename)
+    for filename in get_include_files(filename_spec, symbols):
+        # extend filestack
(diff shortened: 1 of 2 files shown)


src/xacro/__init__.py
```diff
@@ -413,18 +415,27 @@ def process_include(elt, macros, symbols, func):
...
             raise XacroException('namespaces are supported with --inorder option only')
 
-    for filename in get_include_files(filename_spec, symbols):
+    optional = True if optional == "True" else False
+
+    for filename in get_include_files(filename_spec, symbols, optional):
         # extend filestack
         oldstack = push_file(filename)
-        include = parse(None, filename).documentElement
+        try:
```
rhaschke, line 422: Why do you need an additional try-except-block here?
Shouldn't the one in `get_include_files()` be sufficient?

src/xacro/__init__.py
```diff
@@ -362,12 +362,14 @@ def is_include(elt):
...
 
 
-def get_include_files(filename_spec, symbols):
+def get_include_files(filename_spec, symbols, optional=False):
     try:
         filename_spec = abs_filename_spec(eval_text(filename_spec, symbols))
     except XacroException as e:
         if e.exc and isinstance(e.exc, NameError) and symbols is None:
             raise XacroException('variable filename is supported with --inorder option only')
+        elif e.exc.strerror == "No such file or directory" and optional is True:
```
rhaschke: checking for a specific exception type should be done via `isinstance()` as above.

rhaschke (commented):
Thanks for this contribution. See comments below.

test/test_xacro.py
```diff
@@ -1119,6 +1119,36 @@ def test_yaml_support(self):
...
+        src = '''
+<a xmlns:xacro="http://www.ros.org/wiki/xacro">
+  <xacro:include filename="${load_yaml('non-existent.yaml')}"/>
+</a>'''
+        self.assertRaises(xacro.XacroException, self.quick_xacro, src)
+
+    def test_yaml_exist_optional(self):
+        src = '''
+<a xmlns:xacro="http://www.ros.org/wiki/xacro">
+  <xacro:include filename="${load_yaml('non-existent.yaml')}" optional="True"/>
```
rhaschke: Just noticed that this doesn't make sense. `load_yaml()` returns a python dictionary, which is assigned to `filename` here. You need to pass the `optional` argument to `load_yaml`, like this:
`<xacro:property name="dictionary" value="load_yaml('non-existent.yaml', optional=False)"/>`

src/xacro/__init__.py
veimox: We actually need only one try catch, but we need it here because the `.yaml` "existence" evaluation is done in `get_include_files()` whilst the `.xacro` is done in `parse()`. I have made only one try catch now and remove the optional parameter passing to the `get_include_files`

src/xacro/__init__.py
veimox: Done for the new single try-catch

test/test_xacro.py
veimox: As of now i have just put the optional parameter to the `include` attribute. As you suggest is an argument to the function `load_yaml()`. Actually I started doing it like that because my use case is mostly `.yaml` files but that would create two different interfaces for the xacro files and yaml files.

src/xacro/__init__.py
```diff
@@ -431,11 +429,11 @@ def process_include(elt, macros, symbols, func):
             # restore filestack
             restore_filestack(oldstack)
```
rhaschke, line 436: The `filestack` needs to be restored in any case, i.e. also in case of an exception.
Please use `finally` for this purpose.

src/xacro/__init__.py
```diff
@@ -431,11 +429,11 @@ def process_include(elt, macros, symbols, func):
...
 
-        except XacroException as e:
-            if e.exc.strerror == "No such file or directory" and optional is True:
-                continue
-            else:
-                raise
+    except XacroException as e:
+        if e.exc and isinstance(e.exc, IOError) and e.exc.strerror == "No such file or directory" and optional is True:
```
rhaschke: My aim was to avoid the string comparison.

src/xacro/__init__.py
```diff
@@ -413,18 +413,27 @@ def process_include(elt, macros, symbols, func):
...
-        # extend filestack
-        oldstack = push_file(filename)
-        include = parse(None, filename).documentElement
+    optional = True if optional == "True" else False
 
-        # recursive call to func
-        func(include, macros, symbols)
-        included.append(include)
-        import_xml_namespaces(elt.parentNode, include.attributes)
+    try:
```
rhaschke: Do you really want to have the try outside of the for loop?
Consider the case of multiple include files specified with globbing (`*.xacro`). In this case, on a failure, all remaining include files will be skipped as well. Having the try inside the for loop, you can just continue processing the remaining files.

test/test_xacro.py
rhaschke: I'm afraid that I don't really understand your use case. Assigning the result of `load_yaml` (which is a dictionary) to the filename attribute of `xacro:include` doesn't make sense at all to me. Can you provide a real-world use case of yours?

rhaschke (commented):
The PR is still not clear to me.

test/test_xacro.py
veimox: You are right, that test case does not make sense. I did it because it was a easy easy way of triggering a yaml case.

I have dropped now the yaml support because is going through the `property` tag instead of `include` and we can do that in another PR.

src/xacro/__init__.py
veimox: Done

src/xacro/__init__.py
veimox: Done

veimox:
@rhaschke Is there something im missing?

rhaschke:
I rebased onto noetic-devel: #234. I will not introduce new features in older releases anymore.

kota-rr:
> I rebased onto noetic-devel: #234. I will not introduce new features in older releases anymore.

@rhaschke Will this feature not be merged to `melodic-devel`? It will be nice if this can be merged in to melodic as well.

rhaschke:
No, it won't be backported. If you want to use any new feature in Melodic, please just clone the noetic-devel branch into your workspace. The noetic branch is working in Melodic as well.
It's a small project not needing to be compiled, so it's no effort to do this.

PR #234, opened 2020-03-29
optional includes

rhaschke:
Specifying an attribute `optional="True"` for a `xacro:include` tag will silently ignore errors if the file does not exist or is not readable.
This PR is a rebase of #232.


gavanderhoorn:
Was this documented on the wiki?

I cannot find any mention of it on [wiki/xacro](http://wiki.ros.org/xacro).

@veimox: did you intend to document this?

PR #239, opened 2020-05-20
Remove deprecated xacro.py

### LINKED ISSUE #237: `python`: No such file or directory
Via https://github.com/ros/urdf_sim_tutorial/issues/7

With Ubuntu 20.04/noetic, @qboticslabs is having trouble with getting xacro to run Python code. This may go further upstream than `xacro`, but I'm starting here. 

```
$ rosrun xacro xacro.py
/usr/bin/env: ‘python’: No such file or directory
```

sloretz:
Fixes https://github.com/ros/xacro/issues/237

xacro.py is currently installed without using `catkin_install_python()`, so it's shebang isn't properly rewritten. This PR solves the problem by deleting the file; which should be OK since it's deprecated.

I tried installing the file using `catkin_install_python()` before deciding to delete the file, but ran into issues. The file `xacro.py` has some logic to avoid recursively importing itself; but that logic doesn't seem to work in the `devel` space if the file is installed with `catkin_install_python()`. I wasn't able to figure out a way around it.


sloretz:
Deprecated in 2015: https://github.com/ros/xacro/pull/71

sloretz:
@ros-pull-request-builder retest this please

tfoote:
+1 to clear it out, that was deprecated 5 years ago into Jade.

sloretz:
investigating test failures; I [don't see them on the dev job](http://build.ros.org/view/Ndev/job/Ndev__xacro__ubuntu_focal_amd64/3/console) (or locally :() 

Edit: it was caused by me removing `scripts` argument from `setup.py`, same as https://github.com/ros-visualization/rqt_graph/issues/44

sloretz:
@codebot @rhaschke what do you think? If it looks ok; may I merge and release to Noetic?

rhaschke:
Thanks. I missed that during cleaning up...
Could you please add a note in the Noetic migration guide how to adapt roslaunch files?

rhaschke:
release PR: https://github.com/ros/rosdistro/pull/24993

sloretz:
> Could you please add a note in the Noetic migration guide how to adapt roslaunch files?

Added to migration guide: http://wiki.ros.org/noetic/Migration#Use_xacro_instead_of_xacro.py

rhaschke:
@sloretz, thanks.

Issue #240, opened 2020-05-29
Adding "dotted dict" access to yaml loaded dictionaries?


gavanderhoorn:
The yaml support in xacro has been great, it allows for some really flexible `.xacro`s, but one thing always bothered me: with deeply nested dicts, readability really suffers.

Example:

```xml
<limit effort="${settings['my_robot']['my_joint']['limits']['effort']}"
  lower="${settings['my_robot']['my_joint']['limits']['pos_low']}"
  upper="${settings['my_robot']['my_joint']['limits']['pos_up']}"
  velocity="${settings['my_robot']['my_joint']['limits']['vel']}" />
```

A contrived example perhaps, but close to reality.

A work-around has been to extract values and assign them to `xacro:property`s, but that is really more of a work-around and essentially trade readability for having a lot of `xacro:property` lines (with long names).

It would be nice perhaps to access yaml loaded dicts using dotted notation, which would get us:

```xml
<limit effort="${settings.my_robot.my_joint.limits.effort}"
  lower="${settings.my_robot.my_joint.limits.pos_low}"
  upper="${settings.my_robot.my_joint.limits.pos_up}"
  velocity="${settings.my_robot.my_joint.limits.vel}" />
```

still long, but definitely more readable.

Would a PR adding this be accepted into `xacro`?

This doesn't have to be implemented by overriding `__getattr__(..)` on the actual `dict`, but could also be added as an extension of the parser which would split the variable identifier at the `.` and then use the resulting tokens to index into the yaml-loaded dict.

rhaschke:
In such cases as you described, we usually define a local (property) variable:
```xml
<xacro:property name="limits" value="${settings['my_robot']['my_joint']['limits']}" />
```
and then reuse it like this:

```xml
<limit effort="${limits['effort']}" lower="${limits['pos_low']}" upper="limits['pos_up']}" velocity="${limits['vel']}" />
```

But I agree that your suggested notation is more elegant. 
However, the code in `${}` is interpreted directly by python:
https://github.com/ros/xacro/blob/1e3a5b29cb2d97377604cb0a2cc77d08185a8bd4/src/xacro/__init__.py#L525-L527

So I would not mess up further with the (very simple) parser, but indeed overriding `__getattr__()` of the dict returned by `load_yaml()`. In principle, this needs to be done recursively. However, we might want to do it lazily, i.e. only when really accessing a particular sub dict.

rhaschke:
[...]

Issue #236, opened 2020-04-24
xacro give a  ` unknown attribute(s)` warning while include a xacro file with arguments


lianghongzhuo:
xacro give a  ` unknown attribute(s)` warning while include a xacro file with arguments
as discussed in https://github.com/TAMS-Group/tams_ur5/issues/9
we find that in [this line](https://github.com/TAMS-Group/tams_ur5/blob/master/tams_ur5_description/urdf/arm.urdf.xacro#L11) if we include a xacro file with arguments the xacro will give a warning `xacro:include: unknown attribute(s): prefix`.

v4hn:
The unknown attribute is clearly our own bug. The actual error to be reported here, however, lies in the error output of xacro:

```
when processing file: *src/tams/tams_ur5/tams_ur5_description/urdf/ur5.urdf.xacro
included from: */tams/tams_ur5/tams_ur5_description/urdf/arm.urdf.xacro
included from: */src/tams/tams_ur5_setup/tams_ur5_setup_description/urdf/tams_ur5_setup.urdf.xacro
xacro:include: unknown attribute(s): prefix
```

The first line is plain wrong.

The `xacro:include` that triggers the warning is in `arm.urdf.xacro` and the included file is `robotiq-3f-gripper_articulated_macro.xacro`.

`ur5.urdf.xacro` was the last included file in `arm.urdf.xacro` right before the problematic include.
So I guess the processing variables referring to the current file is not updated correctly?

rhaschke:
Maybe, handling errors in xacro:include (which is a corner case) is not yet handled correctly.

rhaschke:
I cannot reproduce the error message. For me, it doesn't show any file context.
So, please provide more information: 

1. Which xacro version you are using?
2. What are your cmdline arguments?

lianghongzhuo:
1 I use the master branch
2. Run `roslaunch tams_u5_bringup tams_ur5.launch` can reproduce the error message.

lianghongzhuo:
Sorry, this is my fault. The error message is for a previous warning for other things, not relevant to the `prefix` bug we introduce.
So I will close this issues.
But maybe we can add some more information when giving `unknown attribute(s):` warning, to tell the user which xacro file has the wrong use of xacro.

rhaschke:
Reopening as a reminder to provide more information for (all) warnings.

PR #245, opened 2020-06-30
[WIP] Allow dotted access to yaml-loaded dicts: d.key1.key2.key3

rhaschke:
Implements #240. Requires some unit test...

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -100,8 +100,11 @@ class YamlDictWrapper(object):
         self.__d = d
 
     def __getattr__(self, item):
-        result = self.__d.__getitem__(item)
-        return YamlDictWrapper(result) if isinstance(result, dict) else result
+        try:
+            result = self.__d.__getitem__(item)
+            return YamlDictWrapper(result) if isinstance(result, dict) else result
+        except KeyError:
+            raise XacroException("No such key: '{}'".format(item))
 
     __getitem__ = __getattr__
 
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -94,6 +94,18 @@ def abs_filename_spec(filename_spec):
     return filename_spec
 
 
+class YamlDictWrapper(object):
+    """Wrapper class providing dotted access to dict items"""
+    def __init__(self, d):
+        self.__d = d
+
+    def __getattr__(self, item):
+        result = self.__d.__getitem__(item)
+        return YamlDictWrapper(result) if isinstance(result, dict) else result
+
+    def __getitem__(self, item):
+        return self.__d.__getitem__(item)
+
 def load_yaml(filename):
     try:
         import yaml
@@ -104,7 +116,7 @@ def load_yaml(filename):
     f = open(filename)
     oldstack = push_file(filename)
     try:
-        return yaml.safe_load(f)
+        return YamlDictWrapper(yaml.safe_load(f))
     finally:
         f.close()
         restore_filestack(oldstack)
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -103,8 +103,7 @@ class YamlDictWrapper(object):
         result = self.__d.__getitem__(item)
         return YamlDictWrapper(result) if isinstance(result, dict) else result
 
-    def __getitem__(self, item):
-        return self.__d.__getitem__(item)
+    __getitem__ = __getattr__
 
 def load_yaml(filename):
     try:
--- a/test/test_xacro.py
+++ b/test/test_xacro.py
@@ -1136,6 +1136,38 @@ class TestXacro(TestXacroCommentsIgnored):
             self.assert_matches(self.quick_xacro(src, cli=['type:=%s' % i]),
                                 res.format(tag=i))
 
+    def test_yaml_support_dotted(self):
+        src = '''
+<a xmlns:xacro="http://www.ros.org/wiki/xacro">
+  <xacro:property name="settings" value="${load_yaml('settings.yaml')}"/>
+  <xacro:property name="type" value="$(arg type)"/>
+  <xacro:include filename="${settings.arms[type].file}"/>
+  <xacro:call macro="${settings.arms[type].macro}"/>
+</a>'''
+        res = '''<a><{tag}/></a>'''
+        for i in ['inc1', 'inc2']:
+            self.assert_matches(self.quick_xacro(src, cli=['type:=%s' % i]),
+                                res.format(tag=i))
+
+    def test_yaml_support_dotted_key_error(self):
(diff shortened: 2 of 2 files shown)


gavanderhoorn:
This implementation basically works, with one 'problem': the following syntax doesn't work:

```xml
<xacro:property name="foo" value="${my_yaml.blop.flip['some_key'].baz}" />
```

Reason this doesn't work is that `flip['some_key']` returns a regular `dict`, not a `YamlDictWrapper`, so the `.` operator doesn't work.

If we feel that is a problem, I can change it, but I wanted to see whether we do (feel that is a problem).

---

Edit: and another issue of course is that anything that's not a valid Python identifier (so something with a space in it fi) or a reserved word will cause problems.

That's a limitation I believe is fair though and we can document.

src/xacro/__init__.py
```diff
@@ -94,6 +94,18 @@ def abs_filename_spec(filename_spec):
...
+    def __init__(self, d):
+        self.__d = d
+
+    def __getattr__(self, item):
+        result = self.__d.__getitem__(item)
+        return YamlDictWrapper(result) if isinstance(result, dict) else result
+
+    def __getitem__(self, item):
+        return self.__d.__getitem__(item)
+
```
gavanderhoorn: Did you have a particular reason to not just `__getitem__ = __getattr__`?

rhaschke: No, I just didn't think about that. This would solve the "problem" you mentioned of course. I'm fine with this change.

gavanderhoorn: I'll check how that works with non-existing keys. I remember there being some difference in the behaviour between `__getitem__` and `__getattr__`.

gavanderhoorn:
@rhaschke: I've submitted ubi-agni/xacro#3 against your fork.

Let me know whether you deem this sufficient.

gavanderhoorn:
Oh, and would you consider this for a backport to Melodic?

gavanderhoorn:
And something else to consider: would we want to make the `YamlDictWrapper` read-only?

I'm not entirely sure whether we could assign something in an expression in `xacro`, but it would be prudent perhaps to explicitly make it impossible in `YamlDictWrapper`.

rhaschke:
Thanks a lot. I will have a look at your contributions over the weekend.

> Would we want to make the YamlDictWrapper read-only?

Isn't it read-only already, because we don't provide `__setitem__` or `__setattr__`?

> Would you consider this for a backport to Melodic?

Yes, in principle no problem. I was fearing some side effects or name conflicts initially. But, as the property return by load_yaml is unique anyway, there should be a problem.

gavanderhoorn:
> > Would we want to make the YamlDictWrapper read-only?
> 
> Isn't it read-only already, because we don't provide `__setitem__` or `__setattr__`?

True.

> > Would you consider this for a backport to Melodic?
> 
> Yes, in principle no problem. I was fearing some side effects or name conflicts initially. But, as the property return by load_yaml is unique anyway, there should be a problem.

That should be "there shouldn't be a problem", correct?

rhaschke:
Yes, sorry. There should**n't** be a problem. :rofl:

gavanderhoorn:
Nice.

Thanks @rhaschke.

rhaschke:
I rebased this onto melodic-devel. Prior to a new release, I want to improve warning messages (#236).

gavanderhoorn:
Ok.

Throwing the `XacroException` on the `KeyError` in the `YamlDictWrapper` makes for quite OK error message though ("no such key 'X', while evaluation expression Y").

But I guess this is more about the general way errors are reported?

gavanderhoorn:
Which branch did this get merged into?

The PR states `noetic-devel`, but I can't find the commits there?

Did it end up in `ubi-agni/xacro` (https://github.com/ubi-agni/xacro/commit/d1171d06d0ebef15eac246ca1b6ef24c3314c9c1)?

rhaschke:
Sorry. Indeed, I missed to push to both, ubi-agni/xacro and ros/xacro. Fixed now.

gavanderhoorn:
Thanks.

Noetic will get it later?

rhaschke:
Yes. I will merge forward changes from Melodic to Noetic after the release. This results in a much cleaner history than, e.g. in MoveIt.

src/xacro/__init__.py
```diff
@@ -104,7 +116,7 @@ def load_yaml(filename):
     f = open(filename)
     oldstack = push_file(filename)
     try:
-        return yaml.safe_load(f)
+        return YamlDictWrapper(yaml.safe_load(f))
```
alikureishy, line 121: Could YamlDictWrapper be exposed for general use in Xacro files? Why restrict the dot notation to just Yaml-loaded files? Why not allow any dictionary to be wrapped with this to enable read-only dotted access? For example:

```
<xacro:property name="dictionary" value="${dict(a=1, b=2}"/>
<xacro:property name="dotted" value="${YamlDictWrapper(dictionary)}"/>
```

However, if I try this directly, I get an error:
` name 'YamlDictWrapper' is not defined`

To make it even more user-friendly, perhaps a function 'dotify' could be exposed as below, that wraps the given dictionary within YamlDictWrapper?
`<xacro:property name="dotted" value="${dotify(dictionary)}"/>
`

rhaschke: That should be possible. Other opinions, @gavanderhoorn, particularly about the function name `dotify`?

gavanderhoorn: Hm.

I'm not sure about the name `dotify`. Seems to be meaningful only if you already know what "dotted access" is.

Can't really come up with anything better right now though.

Also: to avoid breaking existing user's setups: would it make sense to add some sort of prefix? I'm not sure whether that's actually a risk, but seeing as `xacro` is such a common dependency, it would be good to avoid the possibility.

rhaschke: I'm not good at inventing new names either. What kind of prefix do you think of? Something like `xacro.dotify`?
And we warn/throw if the user attempts to override the namespace `xacro`?

gavanderhoorn: > Something like `xacro.dotify`?

yes, that might be nice actually.

> And we warn/throw if the user attempts to override the namespace `xacro`?

that would seem prudent, yes.

rhaschke: I'm hesitant to introduce a `xacro` package namespace. Actually, we already have a safety check issuing a warning if any of our global symbols is overridden: 
https://github.com/ros/xacro/blob/a4f0b84e166de5742f6ac83c561d1e5b92eb3ba7/src/xacro/__init__.py#L325-L326

As overriding works, it will not break existing xacros as well (just issue the warning).

PR #249, opened 2020-08-08
reverting d1171d06d0ebef15eac246ca1b6ef24c3314c9c1

dustingooding:
Removing `YamlDictWrapper`.  `YamlDictWrapper` actively breaks normal Python `dict` functionality and prevents useful functionality like checking to see if a key exists in a `dict`.  `YamlDictWrapper` intended to provide user convenience "dot accessors", but they aren't necessary.  The minor convenience of `foo['bar'].baz` instead of `foo['bar']['baz']` isn't sufficient justification for breaking other useful and valid functionality.

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -94,20 +94,6 @@ def abs_filename_spec(filename_spec):
     return filename_spec
 
 
-class YamlDictWrapper(object):
-    """Wrapper class providing dotted access to dict items"""
-    def __init__(self, d):
-        self.__d = d
-
-    def __getattr__(self, item):
-        try:
-            result = self.__d.__getitem__(item)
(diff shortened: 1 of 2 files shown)


dustingooding:
And please backport this to `melodic`, at least.

rhaschke:
Thanks for providing the unit test, but I'm not going to just revert #245. As you suggested: we need to find another implementation, relying on a `isA-dict` approach.

dustingooding:
You would rather keep an unnecessary dot accessor that's existed for [a month](https://github.com/ros/xacro/commit/d1171d06d0ebef15eac246ca1b6ef24c3314c9c1)... than restore broken functionality that has existed for [five years](https://github.com/ros/xacro/commit/2e49c34fa2e1c37a0e8f21ebae2f1f238a957e72).  I think that's a horrible misunderstanding of priorities.

Issue #251, opened 2020-08-13
yaml: support degrees and radians ctors like rosparam


gavanderhoorn:
`rosparam` supports two convenience functions `rad(..)` and `deg(..)` which are executed before loading parameters which contain them into the parameter server. From [wiki/rosparam](http://wiki.ros.org/rosparam#YAML_Format):

> There are also special converters for angle radian/degree representations. Any Python-legal mathematical expression can be used with the radian value, with pi used to represent pi.
> 
> ```yaml
> angle1: rad(2*pi)
> angle2: deg(180)
> ```

This seems similar to `xacro`s support for "arbitrary Python expressions" in `${}` contexts.

`rosparam` also supports two additional YAML ctors (from the same wiki page):

```yaml
angle1: !degrees 181.0
angle2: !radians 3.14169
```

which result in the same behaviour (ie: values converted to radians before uploading to the parameter server).

If one were to try and load a `.yaml` file using these ctors using `xacro`'s YAML support however, this would not work as `xacro` does not appear to support these:

```
...
could not determine a constructor for the tag '!degrees'
  in "/home/user/catkin_ws/.../joint_limits.yaml", line 18, column 19 
when evaluating expression 'load_yaml(param_file)' 
...
```

This precludes users from re-using the same `.yaml` file for consumption by both `xacro` and `rosparam`.

It would be great if `xacro` could be extended to support the two mentioned YAML constructors.

gavanderhoorn:
PyYAML is configured to use these ctors in `rosparam` [here](https://github.com/ros/ros_comm/blob/12eaf946597739e833b2300ef28bb079f248a9e1/tools/rosparam/src/rosparam/__init__.py#L635-L646).

gavanderhoorn:
Closing as #252 got merged.

PR #252, opened 2020-08-14
Provide support for yaml constructors !degrees and !radians

rhaschke:
Addresses #251. @gavanderhoorn, please review.

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -101,9 +101,30 @@ class YamlDictWrapper(dict):
 
     __getitem__ = __getattr__
 
+
+def construct_angle_radians(loader, node):
+    """utility function to construct radian values from yaml"""
+    value = loader.construct_scalar(node).strip()
+    try:
+        return float(eval(value, global_symbols, global_symbols))
+    except SyntaxError as e:
+        raise XacroException("invalid expression: %s" % value)
+
+
+def construct_angle_degrees(loader, node):
+    """utility function for converting degrees into radians from yaml"""
+    value = loader.construct_scalar(node)
+    try:
+        return math.radians(float(value))
+    except ValueError:
+        raise XacroException("invalid degree value: %s" % value)
+
+
 def load_yaml(filename):
     try:
         import yaml
+        yaml.SafeLoader.add_constructor(u'!radians', construct_angle_radians)
+        yaml.SafeLoader.add_constructor(u'!degrees', construct_angle_degrees)
     except:
         raise XacroException("yaml support not available; install python-yaml")
 
--- a/test/constructors.yaml
+++ b/test/constructors.yaml
@@ -1,0 +1,3 @@
+a: !degrees 180
+b: !radians 0.5*pi
+c: 42
--- a/test/constructors_bad1.yaml
+++ b/test/constructors_bad1.yaml
@@ -1,0 +1,1 @@
+a: !radians os.system('clear')
--- a/test/constructors_bad2.yaml
+++ b/test/constructors_bad2.yaml
@@ -1,0 +1,1 @@
+a: !radians __import__('os').system('clear')
--- a/test/test_xacro.py
+++ b/test/test_xacro.py
@@ -14,6 +14,7 @@ import shutil
 import subprocess
 import re
 import ast
+import math
 try:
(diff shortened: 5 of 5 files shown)


gavanderhoorn:
This is :100: again @rhaschke.

My intention with #251 was not to come up with more work for you, but I'm of course very happy you decided to implement it.

I'll run some tests and let you know.

gavanderhoorn (approved):
This works for me :+1: 

thanks for adding this @rhaschke.

src/xacro/__init__.py
```diff
@@ -101,9 +101,31 @@ def __getattr__(self, item):
...
+        return float(eval(expr))
+    except SyntaxError as e:
+        raise XacroException("invalid expression: %s" % value)
+
+
+def construct_angle_degrees(loader, node):
+    """utility function for converting degrees into radians from yaml"""
+    value = loader.construct_scalar(node)
+    try:
+        return float(value) * math.pi / 180.0
```
gavanderhoorn:
```suggestion
        return math.radians(float(value))
```

gavanderhoorn:
Do I understand the code correctly when I say `!radians` appears to support "arbitrary expressions", while `!degrees` doesn't?

rhaschke:
> Do I understand the code correctly when I say `!radians` appears to support "arbitrary expressions", while `!degrees` doesn't?

Yes, that's what I copied from [ros_param](https://github.com/ros/ros_comm/blob/12eaf946597739e833b2300ef28bb079f248a9e1/tools/rosparam/src/rosparam/__init__.py#L635-L646). Do you want to change this?

gavanderhoorn:
> Yes, that's what I copied from [ros_param](https://github.com/ros/ros_comm/blob/12eaf946597739e833b2300ef28bb079f248a9e1/tools/rosparam/src/rosparam/__init__.py#L635-L646). Do you want to change this?

We should probably keep the same restrictions as in `rosparam`, so I would vote to keep it as-is.

Only thing to make sure perhaps is the `eval(..)` there: does it open up `xacro` to security issues? Should we limit / sanatise something perhaps?

rhaschke:
Indeed, we should restrict `eval()` to our limited set of functions...

src/xacro/__init__.py
```diff
@@ -101,9 +101,30 @@ def __getattr__(self, item):
 
     __getitem__ = __getattr__
 
+
+def construct_angle_radians(loader, node):
+    """utility function to construct radian values from yaml"""
+    value = loader.construct_scalar(node).strip()
+    try:
+        return float(eval(expr, global_symbols, dict(pi=math.pi)))
```
gavanderhoorn:
```suggestion
        return float(eval(value, global_symbols, dict(pi=math.pi)))
```
?

gavanderhoorn: Or `s/value/expr/g` of course.

gavanderhoorn:
@rhaschke: just submitted ubi-agni/xacro#4, which adds some (primitive) testing to the safe(r) `eval(..)` functionality when used in the yaml ctors.

gavanderhoorn:
Thanks @rhaschke :+1:

gavanderhoorn:
@rhaschke: would you have time to release a new version of xacro which includes this change?

rhaschke:
Done: https://github.com/ros/rosdistro/pull/26352

gavanderhoorn:
@rhaschke: I'm not sure I remember your syncing/back-porting policy between branches here.

Would it be possible to get this in Noetic as well?

rhaschke:
Sure. I will prepare a new release later this evening.

rhaschke:
Done: https://github.com/ros/rosdistro/pull/26945

gavanderhoorn:
Thanks @rhaschke :+1:

Briancbn:
Will this be merged into foxy release? I have to merge this to get the melodic-devel-staging urdf models working in ROS2.

rhaschke:
@Briancbn: Sure. Do you mind preparing a merge PR? This would help to speed up the process.

Briancbn:
@rhaschke could you help review the PR? thanks.

rhaschke:
#259 is on my todo list. But there is just too much other work...

PR #253, opened 2020-09-09
Yaml loading: Don't try to wrap a top-level list or scalar as dict

### LINKED ISSUE #254: YamlDictWrapper screws up yaml lists
Hi @gavanderhoorn,
we came across another regression introduced by YamlDictWrapper (#245): A top-level list returned by `load_yaml()` is passed to the dict constructor, which of course fails. Consider this simple xacro example:
```xml
  <xacro:property name="lists" value="${load_yaml('list_of_lists.yaml')}"/>
  ${lists[0]} 
</a>
```
```yaml
- A
- B
```
which fails with: 
```
  File "<stdin>", line 1, in <module>
ValueError: dictionary update sequence element #0 has length 1; 2 is required
```
A simple fix (#253) just ignores non-dicts for wrapping at top-level as well. However, this raises another question: 
What about dicts that are part of lists? Do we want to provide dotted access for those as well? IMO, this would require much more effort, because we would need to provide a `YamlListWrapper` as well to propagate the wrapper down to any element that might get accessed. What do you think? Is that worth the effort?

rhaschke:
Fixes #254.


rhaschke:
Thinking a little more, I came up with a (programming-wise) rather simple solution to handle hierarchical lists and dicts.
However, one should be aware of the fact, that the current implementation copies lists and dicts for each access.

gavanderhoorn:
Just making sure: I'm not opposed to just dropping the whole dotted-dict access.

Against better judgement we tried to make it work, but you've already had to fix two corner cases, which does not seem to bode well for this feature.

rhaschke:
I still think that dotted access is a nice feature. Hopefully, this PR fixes the remaining corner cases :wink:

rhaschke:
Superseeded by #258.

PR #268, opened 2021-03-01
Report correct filename for XML errors

### LINKED ISSUE #267: Error message does not show relevant file
I have a xacro file that includes the other 3 xacro files. One of the included xacro files has an error but the error message could not show which file has the error. 

Here is the main xacro file (robot.xacro)
```xacro

<?xml version="1.0" ?>
<robot xmlns:xacro="http://ros.org/wiki/xacro" name="smart_robot">

. . .
  <xacro:include filename="$(find robot_description)/urdf/robot.base.xacro" />
  <xacro:include filename="$(find robot_description)/urdf/robot.gazebo.xacro" />
  <xacro:include filename="$(find robot_description)/urdf/sensors/rplidar.urdf.xacro" />
 . . . 
 </robot>
```

When I run the xacro command via `rosrun`, this error message has shown on the log
```shell

$ rosrun xacro xacro robot.xacro 
XML parsing error: not well-formed (invalid token): line 5, column 2
when processing file: robot.xacro

Check that:
 - Your XML is well-formed
 - You have the xacro xmlns declaration: xmlns:xacro="http://www.ros.org/wiki/xacro"

```

Adding the verbosity level to the maximum also didn't help.
```shell

$ rosrun xacro xacro --verbosity=4 robot.xacro 
set visualize_sensor: $(arg visualize_sensor) (robot.xacro)
set use_3d_lidar: $(arg use_3d_lidar) (robot.xacro)
set use_arm: $(arg use_arm) (robot.xacro)
set use_stereo: $(arg use_stereo) (robot.xacro)
set tower_base: 0.248 (robot.xacro)
set gpu: $(arg gpu) (robot.xacro)
XML parsing error: not well-formed (invalid token): line 5, column 2
when processing file: robot.xacro

Check that:
 - Your XML is well-formed
 - You have the xacro xmlns declaration: xmlns:xacro="http://www.ros.org/wiki/xacro"
```

The line 5, column 2 in robot.xacro is this line. Which has no error. 
```xacro
<xacro:arg name="use_3d_lidar" default="false"/>
```

Even, removing the line doesn't change the error message. Suddenly, I realize the error was in `robot.gazebo.xacro` file which is included in `robot.xacro` file. Thus, It will be helpful if we can get relevant file path for the error message so we can fix the problem quicker. 

Thank you.

rhaschke:
Fixes #267. Looks like the correct error-location reporting was lost with #234.

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -453,14 +453,13 @@ def process_include(elt, macros, symbols, func):
             func(include, ns_macros, ns_symbols)
             included.append(include)
             import_xml_namespaces(elt.parentNode, include.attributes)
+            # restore filestack
+            restore_filestack(oldstack)
         except XacroException as e:
             if e.exc and isinstance(e.exc, IOError) and optional is True:
                 continue
             else:
                 raise
-        finally:
-            # restore filestack
-            restore_filestack(oldstack)
 
     remove_previous_comments(elt)
     # replace the include tag with the nodes of the included file(s)
--- a/test/test_xacro.py
+++ b/test/test_xacro.py
@@ -959,6 +959,12 @@ class TestXacro(TestXacroCommentsIgnored):
 <a xmlns:xacro="http://www.ros.org/wiki/xacro">
 <xacro:arg name="foo" default=""/>$(arg foo)</a>'''), '''<a/>''')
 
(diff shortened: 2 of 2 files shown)


imaduddinamajid (commented):
LGTM

PR #272, opened 2021-03-31
Scoped macro evaluation [Noetic]

### LINKED ISSUE #270: Macros inside macros are not private once ran
Melodic v1.13.9

**example.xacro**
```xacro
<?xml version="1.0" ?>
<robot name="example" xmlns:xacro="http://www.ros.org/wiki/xacro">
  <xacro:macro name="testMacro" params="">
    <xacro:macro name="testInnerMacro" params="">
    </xacro:macro>
  </xacro:macro>

  <xacro:testMacro/>
  <xacro:testInnerMacro/>
</robot>
```
The above xacro file will succeed when it should not. The `testInnerMacro` should not be available outside of the `testMacro`. The documentation states that "Properties and macros defined within a macro are local to this macro, i.e. not visible outside."

**example2.xacro**
```xacro
<?xml version="1.0" ?>
<robot name="example" xmlns:xacro="http://www.ros.org/wiki/xacro">
  <xacro:macro name="testMacro" params="">
    <xacro:property name="foo" value="bar"/>
  </xacro:macro>

  <xacro:testMacro/>
  <link name="${foo}"/>
</robot>
```
The above file fails as I would expect, because `foo` should not be accessible outside of `testMacro`

rhaschke:
Fixes #270.

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -238,11 +238,17 @@ def eval_extension(s):
 class Table(dict):
     def __init__(self, parent=None):
         dict.__init__(self)
+        if parent is None:
+            parent = dict()  # Use empty dict to simplify lookup
         self.parent = parent
+        try:
+            self.root = parent.root  # short link to root dict / global_symbols
+            self.depth = self.parent.depth + 1  # for debugging only
+        except AttributeError:
+            self.root = parent
+            self.depth = 0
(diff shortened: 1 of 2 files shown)


Doomerdinger (approved):
I am not incredibly farmiliar with the inner workings of xacro, but this all looks ok.
Fixed the issues I was seeing with scoping - but I was only able to test on Melodic.

PR #283, opened 2021-10-01
Expose more functions

rhaschke:
Before exposing even more useful python functions as global symbols, this PR first cleans up the definition of `global_symbols`.
To correctly report the macro call hierarchy via the newly exposed `print_location()` function, I also had to rework the handling of the file and macro stack, which is now hopefully much cleaner.


rhaschke:
Closing + Re-opening to trigger CI

PR #285, opened 2021-10-02
Fix handling of xml namespaces imported via xacro:include in a macro

rhaschke:
Namespace declarations are lifted to the parent of the include node.
In case of a macro instantiation, this is a dummy node whose contents are kept but which is removed itself.
Hence, before removing the dummy node, we need to lift any namespace definitions.


rhaschke:
Replaced by #287.

Issue #280, opened 2021-09-03
[melodic-only] name 'sorted' is not defined in macro argument.


v4hn:
I ran into this trying to run https://github.com/captain-yoshi/scene_parser (as dependency of https://github.com/captain-yoshi/moveit_benchmark_suite) on melodic.

xacro 1.14.8 (noetic) supports this piece of code:
```
<?xml version="1.0" ?>
<robot xmlns:xacro="http://www.ros.org/wiki/xacro" name="test">
  <xacro:macro name="foo" params="sorted_keys:=^">
    <xacro:property name="xyz" value="${sorted_keys}"/>
  </xacro:macro>

  <xacro:foo sorted_keys="${sorted([2,1,0])}"/>
</robot>
```
1.13.12 (melodic) on the other hand fails with

    name 'sorted' is not defined
    when evaluating expression 'sorted([2,1,0])'

At first glance it looks like https://github.com/ros/xacro/pull/278 should have fixed this and was released in 1.13.*12*.
Is the failure in melodic intended or is this an unresolved bug in the branch?

rhaschke:
Actually, it is a bug in Noetic, that `sorted()` works! I'm looking into the issue.

captain-yoshi:
My use case for `sorted` is for looping a map which you can only loop as a list in Xacro (I think). I could change my map to a list but then I loose the ability to get the data by name. Maybe there are other ways to do it without using sorted.

```yaml
apcshelf:
  subframes:
    binA:
      xyz: [0.10479, -0.291029, 1.52393]
    binB:
      xyz: [0.10479, 0, 1.52393]
```

rhaschke:
Sorry, my statement was misleading: I plan to keep `sorted` both in Melodic and Noetic. However, the (current) availability of `sorted` in Noetic pointed me to another issue.
@captain-yoshi: Could elaborate more about your use? What do you mean by "looping a map"? I'm curious how you "unfold" the loop. Can you provide the xacro expression you are using?

v4hn:
> Can you provide the xacro expression you are using?

the full macro I based the minimal example  on can be found [here](https://github.com/captain-yoshi/scene_parser/blob/f5d8801fffc9bebaddae3429a22e92a238b215fc/config/utils.xacro#L25).

rhaschke:
Fixed via 2a130a93adcd593e53d6906958fcf5087ac5d39c.
[...]

PR #291, opened 2021-10-04
`xacro:loop` macro

rhaschke:
As pointed out by @captain-yoshi in https://github.com/ros/xacro/issues/280#issuecomment-912861096 xacro can be used to realize loops:
```xml
<robot name="loop" xmlns:xacro="http://www.ros.org/wiki/xacro">
	<xacro:macro name="loop" params="items:=^">
		<xacro:if value="${items}">
			<!-- pop first item from list -->
			<xacro:property name="item" value="${items.pop(0)}"/>

			<item>${item}</item>

			<!-- recursively call myself -->
			<xacro:loop/>
		</xacro:if>
	</xacro:macro>
	<xacro:loop items="${[1,2,3,4,5]}"/>
</robot>
```

However, it would be even nicer if we could define a generic `loop` macro like this:
```xml
<xacro:macro name="loop" params="items **body:=^">
	<xacro:if value="${items}">
		<!-- pop first item from list -->
		<xacro:property name="item" value="${items[0]}"/>
		<xacro:insert_block name="body" />
		<!-- recursively call myself -->
		<xacro:loop items="${items[1:]}"/>
	</xacro:if>
</xacro:macro>
```
and then just call it with an arbitrary body:
```xml
<xacro:loop items="${python.range(5)}">
	<body>
		<item>${item}</item>
	</body>
</xacro:loop>
```

The attached commits are two (failed) attempts to achieve this. The culprit is the recursive handling of nested macros, which seems to be required to due lazy evaluation of the first generation of xacro.

### DIFF
--- a/examples/loop.xacro
+++ b/examples/loop.xacro
@@ -14,7 +14,7 @@
 
 	<xacro:loop items="${items}">
 		<body>
-			<item>item</item>
+			<item>item: ${item}</item>
 		</body>
 	</xacro:loop>
 </macro_library>
--- a/examples/loop.xacro
+++ b/examples/loop.xacro
@@ -1,0 +1,20 @@
+<?xml version="1.0"?>
+<macro_library xmlns:xacro="http://www.ros.org/wiki/xacro">
+	<xacro:macro name="loop" params="items **body:=^">
+		<xacro:if value="${items}">
+			<!-- pop first item from list -->
+			<xacro:property name="item" value="${items[0]}"/>
+			<xacro:insert_block name="body" />
+			<!-- recursively call myself -->
+			<xacro:loop items="${items[1:]}"/>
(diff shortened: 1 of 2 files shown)


Doomerdinger:
This is interesting, would be a great feature! Many of the issues I've posted to this repo (back to Jan or Feb 2021) were all relating back to some unusual ways I was using xacro, the main one of which was using it to perform looping operations - particularly over arbitrarily sized lists in yaml configurations.

I think in at least one of them I mentioned the desire for this, but I lack the python expertise/familiarity with this repo (and time, mainly time) to try it myself. I ended up doing some other terribleness for speedups (parallelized runs of multiple xacro calls which I stitch together at the end).

If I ever have the time I'll try to look into it more.

PR #303, opened 2021-12-27
convert xacro to pure python package to allow running in Windows

### LINKED ISSUE #302: Cannot run xacro via ros2 run in Windows
```
ros2 run xacro xacro
  File "C:\dev\ros2_rolling\install\Scripts\ros2-script.py", line 33, in <module>
    sys.exit(load_entry_point('ros2cli==0.15.0', 'console_scripts', 'ros2')())
  File "C:\dev\ros2_rolling\install\Lib\site-packages\ros2cli\cli.py", line 89, in main
    rc = extension.main(parser=parser, args=args)
  File "C:\dev\ros2_rolling\install\Lib\site-packages\ros2run\command\run.py", line 70, in main
    return run_executable(path=path, argv=args.argv, prefix=prefix)
  File "C:\dev\ros2_rolling\install\Lib\site-packages\ros2run\api\__init__.py", line 64, in run_executable
    process = subprocess.Popen(cmd)
  File "c:\python38\lib\subprocess.py", line 858, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "c:\python38\lib\subprocess.py", line 1311, in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
OSError: [WinError 193] %1 is not a valid Win32 application
```

this seems like the same error that happened with some other ros2 python programs: https://github.com/ros-visualization/rqt_console/issues/20

I think similar fixes there need to happen to xacro to be able to run this package under windows.

Directly invoking `xacro` also doesnt work since shebangs arent handled by windows.

wmmc88:
closes #302


wmmc88:
Tested locally and running via `ros2 run xacro xacro` now works on windows

rhaschke (commented):
Thanks for providing a potential fix for #302. However, I wasn't expecting such a huge changeset. Switching the build type from `ament_cmake` to `ament_python` is not an option as we provide some cmake functionality to other packages via `xacro-extras.cmake`. Can you think of another option? What is the actual culprit of #302, i.e. what's the difference of installed files?

wmmc88:
> Thanks for providing a potential fix for #302. However, I wasn't expecting such a huge changeset. Switching the build type from `ament_cmake` to `ament_python` is not an option as we provide some cmake functionality to other packages via `xacro-extras.cmake`. Can you think of another option? What is the actual culprit of #302, i.e. what's the difference of installed files?

My understanding is that in windows, running a python script like an executable is just not possible like it is in linux. Linux seems to be able to do this because it interprets the shebang and that's not something that windows does. All of the rqt packages(ie. rqt_graph, rqt_console, rqt_plot, rqt_action, etc. ) solved this issue by converting their packages to a pure python package so that colcon would be able to use the setup.py to create and entrypoint script that is invokable via `ros2 run`. A side effect of this seems to be that none of those packages can be directly invoked anymore... only able to run through `ros2 run` and that seems to be the intended intention by the ros maintainers of those packages.

It seems like `ament_python_install_package` was enhanced recently in https://github.com/ament/ament_cmake/pull/328 and offers a much better solution. new approach in #304.

PR #304, opened 2021-12-28
install xacro using console_scripts entrypoint

### LINKED ISSUE #302: Cannot run xacro via ros2 run in Windows

wmmc88:
This changes xacro to be install using a console_script entrypoint, which now fixes running `xacro` and `ros2 run xacro xacro` in windows. linux functionality should be unchanged. 

closes #302 
supersedes #303


CMakeLists.txt
```diff
@@ -2,17 +2,21 @@ cmake_minimum_required(VERSION 3.5)
...
-install(PROGRAMS scripts/xacro DESTINATION bin)
-install(PROGRAMS scripts/xacro DESTINATION lib/${PROJECT_NAME})
+# install xacro python package w/ console_script entrypoint
+ament_python_install_package(${PROJECT_NAME}
+    SCRIPTS_DESTINATION lib/${PROJECT_NAME})
+
+# copy generated script executables to bin dir so it can be run without ros2 run
+install(
+    DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/ament_cmake_python/${PROJECT_NAME}/scripts/
+    DESTINATION bin/)
```
wmmc88: `ament_python_install_package` now installs the console_script entrypoint so that xacro can be run using `ros2 run xacro xacro`. these lines here make it so that people can still directly call `xacro` since bin is on the PATH but lib isnt

scripts/xacro
```diff
@@ -1,33 +0,0 @@
-#! /usr/bin/env python3
```
wmmc88, line 1: no longer needed. a similar wrapper script (and executable that calls it) is automatically generated by the `setuptools`

setup.py
```diff
@@ -1,35 +0,0 @@
-from setuptools import find_packages
```
wmmc88, line 1: since this ros package wasn't using `ament_python`, I don't think most of this file is ever being properly parsed and used.  at the very least `data_files` and `entry_points` isn't being handled and all that functionality here is already taken care of in the CMakelists so I thought I should remove this misleading file. All of this info is already duplicated elsewhere.

rhaschke (commented):
Thanks. This looks more promising!
~~I was thinking about [symlinking](https://stackoverflow.com/questions/35765106/symbolic-links-cmake) from `lib/xacro` to `bin` instead of installing the file a second time.~~
Symlinks are not supported in Windows.

rhaschke:
Looks like the new feature was only released into `rolling` so far. The backport to galactic [was merged on Dec 8th](https://github.com/ament/ament_cmake/pull/364), but [not yet released](https://github.com/ament/ament_cmake/tags).

wmmc88:
> Looks like the new feature was only released into `rolling` so far. The backport to galactic [was merged on Dec 8th](https://github.com/ament/ament_cmake/pull/364), but [not yet released](https://github.com/ament/ament_cmake/tags).

Ah i didnt notice this because I use rolling. Does this change need to wait until backports to non-eol distros are released then?

CMakeLists.txt
```diff
@@ -2,17 +2,16 @@ cmake_minimum_required(VERSION 3.5)
...
+ament_python_install_package(${PROJECT_NAME} SCRIPTS_DESTINATION lib/${PROJECT_NAME})
 
-# install xacro python package
-ament_python_install_package(xacro)
-# install main script
-install(PROGRAMS scripts/xacro DESTINATION bin)
-install(PROGRAMS scripts/xacro DESTINATION lib/${PROJECT_NAME})
+# install entry-point script(s) in bin as well
+install(PROGRAMS ${CMAKE_CURRENT_BINARY_DIR}/ament_cmake_python/${PROJECT_NAME}/scripts/xacro
+        DESTINATION bin)
```
wmmc88: Using `PROGRAMS` instead breaks the build in windows:
```
CMake Error at cmake_install.cmake:68 (file):
  file INSTALL cannot find
  "C:/Users/melvi/git-repos/ros2_upstream_ws/build/xacro/ament_cmake_python/xacro/scripts/xacro":
  File exists.
  ```
  In windows, the generated file is `xacro.exe`, not `xacro`

rhaschke: Argh. Hopefully the next commit fixes this.

rhaschke:
> Does this change need to wait until backports to non-eol distros are released then?

Yes, I cannot release in this state. I also suggested some changes: https://github.com/ament/ament_cmake/pull/328#issuecomment-1002522954

Tobias-Fischer:
I think the backport has been released. Could this PR be merged now?

rhaschke:
I think this is also blocked by https://github.com/ament/ament_cmake/pull/372. Will double check...

rhaschke:
Closing and reopening to trigger CI.

PR #318, opened 2022-06-09
Dotted YAML access from list iterator

### LINKED ISSUE #317: load_yaml dot accessor breaks down in loops
I've notice that if I call (from python code, not within a xacro file) `load_yaml` the dot accessor does not work in for loops.

Given this yaml file:
```yaml
---
list_ex:
  - name: foo
    val:
      x: 1
  - name: bar
    val:
      x: 2
```

The following will not work correctly:
```python
b = 0
loaded_yaml = load_yaml(file)
for o in loaded_yaml.list_ex:
    b = b + o.val.x
```
It will complain that `'dict' has no value 'val'`

However, the following will work fine
```python
b = 0
loaded_yaml = load_yaml(file)
for x in range(0, len(loaded_yaml.list_ex)):
    o = loaded_yaml[x]
    b = b + o.val.x
```

Calling this code directly is not the intended use case, so this may not be worth addressing.

rhaschke:
Fixes #317


Doomerdinger:
Works for me.

PR #323, opened 2023-05-18
Updated error type in YamlDictWrapper

Doomerdinger:
This allows the `hasattr` function to work properly.

Also changed the error message to be more similar to the standard error message for attribute access.

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -110,7 +110,7 @@ class YamlDictWrapper(dict):
         try:
             return YamlListWrapper.wrap(super(YamlDictWrapper, self).__getitem__(item))
         except KeyError:
-            raise XacroException("No such key: '{}'".format(item))
+            raise AttributeError("'YamlDictWrapper' object has no attribute '{}'".format(item))
 
     __getitem__ = __getattr__
 
--- a/test/test_xacro.py
+++ b/test/test_xacro.py
@@ -1514,6 +1514,11 @@ included from: string
 </a>'''
         self.assertRaises(xacro.XacroException, self.quick_xacro, src.format(file=file))
 
+    def test_yaml_hasattr_support(self):
+        yaml = xacro.load_yaml('settings.yaml')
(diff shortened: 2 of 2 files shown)


rhaschke:
Could you please explain why `hasattr` didn't work as expected. Ideally, provide a new unit test that fails without your fix.

Doomerdinger:
Sure.
`hasattr` does not work correctly because of its underlying implementation.
Essentially, `hasattr` will call `getattr` and if `getattr` raises an `AttributeError` it catches that and returns `False`. If it does not raise an `AttributeError` it will return true.
By raising an error besides `AttributeError` in the `getattr` function, instead of returning true or false we get an exception instead.

I could make a new xacro exception type that inherits from `AttributeError` so that `hasattr` will still work, but we get a custom xacro exception type.

I will add a test.

src/xacro/__init__.py
```diff
@@ -110,7 +110,7 @@ def __getattr__(self, item):
         try:
             return YamlListWrapper.wrap(super(YamlDictWrapper, self).__getitem__(item))
         except KeyError:
-            raise XacroException("No such key: '{}'".format(item))
+            raise AttributeError("'YamlDictWrapper' object has no attribute '{}'".format(item))
```
rhaschke, line 113:
```suggestion
            raise AttributeError("The yaml dictionary has no key '{}'".format(item))
```

rhaschke (approved):
Thanks for the clarification. I generally approve with a minor rewording of the error msg.
Could you please rebase against melodic-devel. I will later merge-forward into noetic-devel.

src/xacro/__init__.py
Doomerdinger: Yes, will do.

As a note:
The implementation for python2 and python3 varies slightly here for `hasattr`, but the suggested  implementation should work correctly for both.

Doomerdinger:
New PR (melodic) [here](https://github.com/ros/xacro/pull/324).

PR #324, opened 2023-05-22
Updated error type in YamlDictWrapper

Doomerdinger:
This allows the `hasattr` function to work properly.

Also changed the error message to be more similar to the standard error message for attribute access.

Further context [here](https://github.com/ros/xacro/pull/323).


rhaschke:
Thanks for rebasing!

PR #329, opened 2024-01-03
Installation and setup without ROS

hubernikus:
I added a `pyproject.toml` and a `requirements.txt` file which enable ROS independent installation. I required this on my system to get it working, and use the xacro library to create URDF-models of robots.

Installation can then be done with:
```
pip install git+https://github.com/hubernikus/xacro.git@ros2
```


c-rizz:
Hi, I was dealing with the same problem and tried your fork, however it still fails because of the ament_index_python import in xacro/substitution_args.py, which I think is not available on pypi. This could be mostly solved by moving the import inside _eval_find() at line 141. Then it would only fail if the import is actually needed.

hubernikus:
Interesting that did not trigger for me. 

But yes either moving it in the _eval_find() or put it in a try / except statement would resolve it.

rhaschke:
`substitution_args` and thus `ament_index_python` are imported conditionally only when _using_ substitution args:
https://github.com/ros/xacro/blob/7231b8bd7a7c8227d2234702e20fb49c598306c0/xacro/__init__.py#L287-L292

rhaschke:
I cleaned your PR a little bit:
- requirements.txt is not needed as you already specify `dependencies` in `pyproject.toml`.
- fixed version information
- added console script generation

PR #330, opened 2024-01-05
Add support for devcontainers

BrunoB81HK:
Hi!

I found myself using `xacro` quite a lot recently and I wanted to try to add some features as a mean to get familiar with its inner working. One way to make it easier to develop is to use [devcontainers](https://code.visualstudio.com/docs/devcontainers/containers), so I created some that works for recent ROS 2 distros. It can probably be improved, but I think this is a good starting point.

Let me know what you think!

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,28 @@
+# Set base image
+ARG ROS_DISTRO=humble
+FROM ros:${ROS_DISTRO}-ros-base
+
+# Arguments
+ARG GID=1000
+ARG UID=1000
+
+# Install some dependencies packages
+RUN apt update -q \
+ && apt upgrade -q -y \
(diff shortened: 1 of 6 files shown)


rhaschke (commented):
I am using devcontainers too, but I don't commit the config files into a package's source tree. Rather I use generic (or existing) ROS containers for several projects.

BrunoB81HK:
Hi @rhaschke,

I agree that most of the default ones are good enough for this repo, although I think a post-create script could be useful to install pre-commit and run `colcon build --symlink-install` so that the workspace is ready once opened.

rhaschke:
You can make all of this part of your docker image, can't you? I don't see the need to define something like this for a specific source repo.

BrunoB81HK:
Indeed you can. I think you are right, with the defaults available, this is not really necessary.

PR #333, opened 2024-01-24
Add python's builtin abs() function

### LINKED ISSUE #332: Feature request: abs() function in ${.} expressions
While implementing some too complicated calculations in xacro I noticed that python's `abs` is not usable in xacro.
Sorry if it is available, but I could not find it in the `python` nor in the `math` namespace, and I could not find it in any documentation.

mfbehrens:
Fixed #332.

### DIFF
--- a/src/xacro/__init__.py
+++ b/src/xacro/__init__.py
@@ -197,12 +197,12 @@ def create_global_symbols():
 
     deprecate_msg = 'Using {name}() directly is deprecated. Use {ns}.{name}() instead.'
     # This is the list of symbols we have exposed for years now. Continue exposing them directly
-    expose('list', 'dict', 'map', 'len', 'str', 'float', 'int', 'True', 'False', 'min', 'max', 'round',
+    expose('list', 'dict', 'map', 'len', 'str', 'float', 'int', 'True', 'False', 'min', 'max', 'round', 'abs',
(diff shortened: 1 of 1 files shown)


src/xacro/__init__.py
```diff
@@ -197,12 +197,12 @@ def expose(*args, **kwargs):
 
     deprecate_msg = 'Using {name}() directly is deprecated. Use {ns}.{name}() instead.'
     # This is the list of symbols we have exposed for years now. Continue exposing them directly
-    expose('list', 'dict', 'map', 'len', 'str', 'float', 'int', 'True', 'False', 'min', 'max', 'round',
+    expose('list', 'dict', 'map', 'len', 'str', 'float', 'int', 'True', 'False', 'min', 'max', 'round', 'abs',
            source=__builtins__)
```
rhaschke, line 201: I would like to avoid cluttering the global namespace with even more names.

rhaschke (approved):

src/xacro/__init__.py
rhaschke: Implemented via d9b66790de780fb125b5b3d7fbae9b2139559ca0

rhaschke:
Fixed in d9b66790de780fb125b5b3d7fbae9b2139559ca0

PR #336, opened 2024-02-23
Escape regexes using raw strings.

adamheins:
This fixes some warnings about invalid escape sequences, which are indeed not intended to be interpreted as escape sequences but rather as regex sequences.

### DIFF
--- a/xacro/__init__.py
+++ b/xacro/__init__.py
@@ -555,8 +555,8 @@ def is_valid_name(name):
     return False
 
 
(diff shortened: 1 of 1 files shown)


rhaschke:
Thanks.

ruffsl:
Thank you! This was annoying to trackdown in Iron, but glad to see it's fixed already upstream with v2.0.11-1in Noble.

rhaschke:
Looks like I missed a release into Iron. Added now: https://github.com/ros/rosdistro/pull/42712.

PR #350, opened 2024-12-18
Create xacro Bazel module

mjcarroll:
Add the infrastructure necessary to use xacro as a bazel module.  This gives users the ability to either use xacro as a standalone executable in their bazel projects, but also adds rules to allow for bazel to run xacro at build-time.

This also adds bazel-based CI to validate that the workflows continue to work.


mjcarroll:
@rhaschke I talked with @codebot about this addition, and he seemed generally inclined to take it.  For context, we use the bazel build system and xacro together currently, but would like to land it in the bazel central registry (https://github.com/bazelbuild/bazel-central-registry) so that people can quickly add xacro to their bazel projects.

This is also currently being done in drake (https://github.com/RobotLocomotion/drake/tree/master/tools/workspace/ros_xacro_internal), so maybe an opportunity to condense to a single set of rules (@jwnimmer-tri)

mjcarroll:
Note that this is still wip as I haven't gotten the tests running under bazel yet, but wanted to start the conversation.

jwnimmer-tri:
Thanks for the ping!

Yes, if xacro were published in the bazel central registry then Drake would probably switch to using that module definition (assuming it works equally well to our current solution).

If I can help out (e.g., if there are questions about how bazel works), just let me know.

rhaschke:
I never used bazel before. Are there plans to bazalize a large set of ROS2 packages or is this only for xacro?
When including bazel build rules, we should also have a CI test validating them.

mjcarroll:
> Are there plans to bazalize a large set of ROS2 packages or is this only for xacro?

I don't have any plans to do more than `xacro` at the moment. We mostly want to be able to run it at build time in `bazel` so that it integrates cleanly.

There is recurring interest in the broader ROS community for bazel and some community supported repos (https://github.com/mvukov/rules_ros, https://github.com/ApexAI/rules_ros) as well as Bazel users in adjacent projects like Drake and Gazebo.

I'm happy to find a reviewer who is familiar and I will add CI as well.

mjcarroll:
Following up to add CI: https://github.com/ros/xacro/pull/351

rhaschke (commented):
Can we avoid cluttering the root folder with all those bazel files?
Is it possible to move them into a subfolder?

mjcarroll:
> Can we avoid cluttering the root folder with all those bazel files?

I moved all of the ones possible into a subdirectory.  MODULE.bazel and BUILD.bazel files have to appear at the root, much like a CMakeLists.txt, so they aren't possible to move.

The .bazelversion file helps with automatic version tooling like https://github.com/bazelbuild/bazelisk and the CI workflow that I'm using here.  It must also appear in the root, but since it's a dotfile, I'm hoping is less of an inconvenience?

mjcarroll:
@jwnimmer-tri I think I'm pretty happy with the rules and tests that I have created here, mind to take a look from a bazel perspective?

I have also attempted to replace the drake impl here: https://github.com/RobotLocomotion/drake/pull/22341

BUILD.bazel
```diff
@@ -0,0 +1,80 @@
...
+)
+
+py_library(
+    name = "xacro_lib",
+    srcs = glob(
+        [
+            "xacro/**/*.py",
+        ],
+        allow_empty = False,
+    ),
```
mbeards: Does this really need to be a glob?

bazel/defs.bzl
```diff
@@ -0,0 +1,135 @@
...
+)
+
+def _xacro_impl(ctx):
+    if ctx.outputs.out:
+        out = ctx.outputs.out
+    else:
+        src = ctx.file.src.basename
+        if not src.endswith(".xacro"):
+            fail("xacro_file src should be named *.xacro not {}".format(src))
+        out = ctx.actions.declare_file(src[:-6])
```
mbeards: Derive this from the length of ".xacro" for readability.

.github/workflows/bazel.yml
```diff
@@ -0,0 +1,21 @@
+name: Bazel CI
+on: [push, pull_request, workflow_dispatch]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
```
mering: Nit: IIUC this is also used for post merge (`on.push` for `main`). With `cancel-in-progress` you are loosing the signal on which commit might have broken `main` as some of the workflows might get cancelled when multiple PRs are merged quickly.

bazel/defs.bzl
```diff
@@ -0,0 +1,135 @@
...
+
+    return [
+        XacroInfo(result = out, data = xacro_data),
+        DefaultInfo(
+            files = depset([out]),
+            data_runfiles = ctx.runfiles(files = [out]),
+        ),
+    ]
+
+_xacro_rule = rule(
```
mbeards: Add `provides = [XacroInfo, DefaultInfo]`

.github/workflows/bazel.yml
```diff
@@ -0,0 +1,21 @@
...
+    with:
+      folders: |
+        [
+          ".",
+          "bazel/integration_test",
+        ]
+      exclude: |
+        [
+          {"folder": ".", "bzlmodEnabled": false},
+          {"folder": "bazel/integration_test", "bzlmodEnabled": false},
```
mering, line 20: I haven't looked at the rules bug `bzlmodEnabled` sounds not like it is testing Bzlmod support. Consider adding a comment explaining why this needs to be like this.

bazel/defs.bzl
```diff
@@ -0,0 +1,135 @@
...
+        "_xacro": attr.label(
+            default = "@xacro//:xacro",
+            cfg = "host",
+            executable = True,
+        ),
+    },
+    implementation = _xacro_impl,
+)
+
+def xacro_file(
```
mbeards: Why define this macro here vs just exposing the rule directly?

MODULE.bazel
```diff
@@ -0,0 +1,21 @@
...
+    python_version = "3.11",
+)
+
+pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
+pip.parse(
+    hub_name = "pypi",
+    python_version = "3.11",
+    requirements_lock = "//bazel:requirements_lock.txt",
+)
+use_repo(pip, "pypi")
```
mering: Consider adding support for multiple Python versions: https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/grpc/1.68.0/MODULE.bazel#L49-L78.

mering: The repo name needs to be unique among all modules. So using `pypi` is not a good choice. Prefer something like `xacro_python_dependencies`.
```suggestion
    hub_name = "xacro_python_dependencies",
    python_version = "3.11",
    requirements_lock = "//bazel:requirements_lock.txt",
)
use_repo(pip, "xacro_python_dependencies")
```

bazel/defs.bzl
mjcarroll: Follow drake's implementation, but there isn't a strong reason to not just expose the rule, so dropped this macro.

BUILD.bazel
mjcarroll: No, removed.

bazel/defs.bzl
mjcarroll: Done.

bazel/defs.bzl
mjcarroll: Done.

MODULE.bazel
mjcarroll: Done, thanks.

mjcarroll: Good catch.

.github/workflows/bazel.yml
mjcarroll: These are skipping tests with bzlmod disabled  (that is, only testing with bzlmod), which is invoked here: https://github.com/bazel-contrib/.github/blob/master/.github/workflows/bazel.yaml#L121

At this point, I don't think it makes sense to add support for workspaces.  I will add a note.

.github/workflows/bazel.yml
mjcarroll: Removed, thanks.

mering: For pull requests this is usually desired though. Also note that a push to a PR branch will always create two workflow runs (push and pull_request). So I usually create a workflow with on.workflow_call and then separate entrypoints for pull request and push to default branch.

.github/workflows/bazel.yml
mering: I might have a logical twist somewhere but I read the linked line as: If `bzlmodEnabled==true` then `--enable_bzlmod=true --enable_workspace=false`. So false is the opposite which we don't seem to desire.

rhaschke (commented):
I don't yet understand the purpose of `open_bazel`. Is that to replace `$(find pkg)` with a search in the `XACRO_INPUTS` environment variable? If so, why not make it more explicit and introduce a new find directive, e.g. `$(find bazel)`.
In any case, this feature needs to be better documented.

mjcarroll:
> I don't yet understand the purpose of open_bazel. Is that to replace $(find pkg) with a search in the XACRO_INPUTS environment variable?

Since bazel does all builds in a hermetically-sealed workspace (essentially chroot), you can't refer to the files with their absolute filesystem paths.  Instead, you have to pre-declare in the build the data files you intend to use and then refer to them via their bazel paths (prefixed with `//` in same repo or `@` across repo).

I wouldn't be opposed to adding a bazel-specific directive, though.

.github/workflows/bazel.yml
mjcarroll: > I might have a logical twist somewhere 

This is in the "exclude" section.  So we are excluding the test configurations where bzlmodEnabled=false.  I agree that it's confusing.

mering: Thanks for the clarification!

mbeards:
> > I don't yet understand the purpose of open_bazel. Is that to replace $(find pkg) with a search in the XACRO_INPUTS environment variable?
> 
> Since bazel does all builds in a hermetically-sealed workspace (essentially chroot), you can't refer to the files with their absolute filesystem paths. Instead, you have to pre-declare in the build the data files you intend to use and then refer to them via their bazel paths (prefixed with `//` in same repo or `@` across repo).
> 
> I wouldn't be opposed to adding a bazel-specific directive, though.

Does the absolute path of the input files end up baked into the output? If not, we could make the invoking rule copy or symlink the inputs into a single directory and invoke `xacro` on those files, rather than trying to make `xacro`'s internal path resolution support Bazel's semantics.

(I think in the gRPC logic https://github.com/grpc/grpc/pull/27275/files we're doing something ~ similar to handle x-repo Python codegen.)

mjcarroll:
> Does the absolute path of the input files end up baked into the output? If not, we could make the invoking rule copy or symlink the inputs into a single directory and invoke xacro on those files, rather than trying to make xacro's internal path resolution support Bazel's semantics.

No, the absolute path shouldn't show up anywhere in the output, I'll give this approach a shot.

mjcarroll:
@rhaschke I have updated my implementation here to remove any bazel-specific semantics that were there before.  The small change I did need to make was an optional `--root-dir` flag to allow users to specify where the included xacro files should be resolved from.  I have also added tests so this is exercised from the non-bazel paths in addition to the bazel workflows testing it.

xacro/__init__.py
```diff
@@ -1018,7 +1022,8 @@ def parse(inp, filename=None):
     f = None
     if inp is None:
         try:
-            inp = f = open(filename)
+            global root_dir
```
rhaschke: You are just reading `root_dir`. There is no need to declare the variable global here, right?
```suggestion
```

rhaschke (commented):
Thanks for cleaning this up.
Why do you need so many integration-test files? What exactly are they testing?
As I understand, the unit tests in `test_xacro.py` are run as well, aren't they?

mjcarroll:
> Why do you need so many integration-test files? What exactly are they testing?

They are testing that the bazel rules correctly can find/expand xacro files as part of a build.  It likely looks like more files because it's an entirely separate bazel project.  The validates that there isn't anything "leaking" between the root bazel module implementation.

> As I understand, the unit tests in test_xacro.py are run as well, aren't they?

That is correct, the stuff in the integration test only tests the bazel rules, similar to the cmake variant in the `test` directory.

PR #351, opened 2024-12-19
Add bazel CI workflow

mjcarroll:


mjcarroll:
This is still expected to fail until I add bazel-specific tests.

.github/workflows/ci.bazelrc
```diff
@@ -0,0 +1,18 @@
+# This file contains Bazel settings to apply on CI only.
+# It is referenced with a --bazelrc option in the call to bazel in ci.yaml
```
rhaschke, line 2: I cannot find this reference. Can you please point me to the line in code?

rhaschke (commented):
What's the purpose of the folder `test/bazel`? 
If that's just the build folder for bazel, I suggest to move it to `.github/workflows/bazel`.

rhaschke:
Please integrate this PR into #350. They build upon each other anyway.

mjcarroll:
> What's the purpose of the folder test/bazel?

It's where an integration test of the bazel rules will go, I haven't implemented it yet.   I will move it into `bazel/integration_test` as part of cleaning up the bazel-specific files.

.github/workflows/ci.bazelrc
mjcarroll: Well, it used to be used as part of the upstream CI workflow, but when trying to find the reference, I find that it has been removed in https://github.com/bazel-contrib/.github/pull/11/files#diff-fcdc74e04f54e030b7b5cfde9bbcd1cc443162841efd9375481872e030b43a2bL153, which was reported upstream here: https://github.com/bazel-contrib/.github/issues/23)

PR #353, opened 2025-01-08
Python 3.13 introduced new attr argument to _write_data()

### LINKED ISSUE #352: Internal xml minidom API changed in Python 3.13 (_write_data)
Python 3.13 has merged gh-50002, which changes the API of the internal _write_data function ([issue](https://github.com/python/cpython/issues/50002), [PR](https://github.com/python/cpython/pull/107947)).

Used here:
https://github.com/ros/xacro/blob/e956f945a31901773801dafe90f5aa4905b0f981/xacro/xmlutils.py#L123

The new parameter "attr" seem to enable some additional escaping for attributes. Looks like an attribute is being written here, so should probably be set to true (or better yet, find a replacement in the public API).

rhaschke:
Hopefully fixes #352.
@oysstu Could you please review and validate on Python 3.13?

### DIFF
--- a/src/xacro/xmlutils.py
+++ b/src/xacro/xmlutils.py
@@ -106,6 +106,13 @@ def reqd_attrs(tag, attrs):
             raise RuntimeError("%s: missing attribute '%s'" % (tag.nodeName, name))
     return result
 
+# Python 3.13 introduced a new argument to xml.dom.minidom._write_data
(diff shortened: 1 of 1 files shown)


oysstu:
Yes, that seems to fix the problem (tested on ros2 branch)

PR #355, opened 2025-02-05
bazel: bump version number to match package.xml

mjcarroll:


rhaschke:
Is there a way to automatically fetch the version number from package.xml? I will typically forget bumping the bazel version number when doing a release...

mjcarroll:
Actually, I have an alternative, closing this out.

PR #357, opened 2025-02-13
bazel: remove version number from MODULE.bazel

mjcarroll:
This is unneeded and injected at the central registry.
This helps eliminate maintainer churn around keeping this in sync.

Replaces: #355


rhaschke (approved):
I don't yet understand how the version number is injected, but I just believe you.

mjcarroll:
It will happen in a patch in BCR: https://github.com/bazelbuild/bazel-central-registry/pull/3731

I did this for the gazebo packages as well, and forgot that it was easier to just exclude the version from the upstream to decrease maintenance burden.

PR #358, opened 2025-02-14
bazel: Correct path logic for input files in third-party modules

mjcarroll:
When integrating with an external bazel module, some of the path logic was incorrect.

This updates the logic for the bazel rules to correctly account for this case.


rhaschke (approved):
Approving xacro/__init__.py. Didn't look at bazel-related files. Please squash merge.

PR #360, opened 2025-04-26
Use Label() to support different repo_name

mering:
Resolve `xacro` label correctly when used with a different `repo_name`. This can be done by using `Label()` to resolve the label relative to the `.bzl` file instead of relative to the `BUILD` file where the macro is called.


mjcarroll (approved):

rhaschke:
@mering, thanks for the fix and @mjcarroll thanks for reviewing.

PR #370, opened 2025-08-06
Expose bool in expressions

### LINKED ISSUE #369: Allow `builtins.bool` in xacro expressions
The wiki states that `bool` is an available function in xacro expressions.
However, this is not the case.

A use case for this seemingly trivial function is to convert things to check if parameters are truethy (for `list` and `str` non-empty, for `int` non-zero, etc.).

This could then be used in `xacro:if` statements which require a boolean value.

Was `bool` omitted from the permitted functions on purpose?

----
If it was on purpose, then the GitHub wiki page needs to be updated.
Otherwise it needs to be added to the exposed functions.

I could make a pull request for either option.

SuperJappie08:
Fixes #369

### DIFF
--- a/xacro/__init__.py
+++ b/xacro/__init__.py
@@ -202,9 +202,9 @@ def create_global_symbols():
     # This is the list of symbols we have exposed for years now. Continue exposing them directly
     expose('list', 'dict', 'map', 'len', 'str', 'float', 'int', 'bool', 'True', 'False', 'min', 'max', 'round',
            source=__builtins__)
-    # These few were only recently added. The should move into python namespace, but (with a deprecation msg) stay global for now
+    # These few were only recently added. They should move into the python namespace, but (with a deprecation msg) stay global for now
     expose('sorted', 'range', source=__builtins__, ns='python', deprecate_msg=deprecate_msg)
-    # Expose all builtin symbols into the python namespace. Thus the stay accessible if the global symbol was overriden
+    # Expose all builtin symbols into the python namespace. Thus, they stay accessible if the global symbol is overridden
     expose('list', 'dict', 'map', 'len', 'str', 'float', 'int', 'bool', 'True', 'False', 'min', 'max', 'round',
            'abs', 'all', 'any', 'complex', 'divmod', 'enumerate', 'filter', 'frozenset', 'hash', 'isinstance', 'issubclass',
            'ord', 'repr', 'reversed', 'slice', 'set', 'sum', 'tuple', 'type', 'vars', 'zip', source=__builtins__, ns='python')
--- a/xacro/__init__.py
+++ b/xacro/__init__.py
@@ -200,12 +200,12 @@ def create_global_symbols():
 
     deprecate_msg = 'Using {name}() directly is deprecated. Use {ns}.{name}() instead.'
     # This is the list of symbols we have exposed for years now. Continue exposing them directly
-    expose('list', 'dict', 'map', 'len', 'str', 'float', 'int', 'True', 'False', 'min', 'max', 'round',
+    expose('list', 'dict', 'map', 'len', 'str', 'float', 'int', 'bool', 'True', 'False', 'min', 'max', 'round',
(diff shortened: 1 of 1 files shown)


xacro/__init__.py
```diff
@@ -200,12 +200,12 @@ def expose(*args, **kwargs):
 
     deprecate_msg = 'Using {name}() directly is deprecated. Use {ns}.{name}() instead.'
     # This is the list of symbols we have exposed for years now. Continue exposing them directly
-    expose('list', 'dict', 'map', 'len', 'str', 'float', 'int', 'True', 'False', 'min', 'max', 'round',
+    expose('list', 'dict', 'map', 'len', 'str', 'float', 'int', 'bool', 'True', 'False', 'min', 'max', 'round',
            source=__builtins__)
     # These few were only recently added. The should move into python namespace, but (with a deprecation msg) stay global for now
```
rhaschke:
```suggestion
    # These few were only recently added. They should move into the python namespace, but (with a deprecation msg) stay global for now
```

xacro/__init__.py
```diff
@@ -200,12 +200,12 @@ def expose(*args, **kwargs):
...
     expose('sorted', 'range', source=__builtins__, ns='python', deprecate_msg=deprecate_msg)
     # Expose all builtin symbols into the python namespace. Thus the stay accessible if the global symbol was overriden
```
rhaschke:
```suggestion
    # Expose all builtin symbols into the python namespace. Thus, they stay accessible if the global symbol is overridden
```

xacro/__init__.py
```diff
@@ -200,12 +200,12 @@ def expose(*args, **kwargs):
 
     deprecate_msg = 'Using {name}() directly is deprecated. Use {ns}.{name}() instead.'
     # This is the list of symbols we have exposed for years now. Continue exposing them directly
-    expose('list', 'dict', 'map', 'len', 'str', 'float', 'int', 'True', 'False', 'min', 'max', 'round',
+    expose('list', 'dict', 'map', 'len', 'str', 'float', 'int', 'bool', 'True', 'False', 'min', 'max', 'round',
```
rhaschke, line 203: Please omit from this list, which exposes the symbols into the _global_ namespace.
```suggestion
    expose('list', 'dict', 'map', 'len', 'str', 'float', 'int', 'True', 'False', 'min', 'max', 'round',
```

rhaschke (commented):
Thanks for this PR. I don't remember a reason to omit `bool` from the list of exposed functions. Let's take the chance to fix some typos as well ;-)

xacro/__init__.py
SuperJappie08: The way it's currently phrased on the [wiki](https://github.com/ros/xacro/wiki#available-functions-and-symbols), it explicitly states that it is in the global namespace .

I'm fine with removing it, but I thought it made sense since it is a conversion function like `int`, `float` and `str`.

rhaschke: Formally, the symbol was not exposed globally in the past and therefore should not be included in this list. Anyway, I agree with your argumentation: it is more convenient to use `bool()` directly, instead of `python.bool()`.
Given that `bool` was probably not used as a property variable name frequently, I'm also fine adding it to the global namespace.

rhaschke:
Do you mind retargeting (and rebasing) your PR onto the `noetic-devel` branch, which is still the main development branch of this repo. I will merge the changes into the ros2 branch later on.

SuperJappie08:
> Do you mind retargeting (and rebasing) your PR onto the `noetic-devel` branch, which is still the main development branch of this repo. I will merge the changes into the ros2 branch later on.

No problem, will do

SuperJappie08:
It will be in a new PR if you don't mind, since I didn't branch nicely. (It would require changing my `ros2` branch to be equivalent to `noetic-devel`, which seems more complex than necessary)

SuperJappie08:
Closing in favor of #371

PR #371, opened 2025-08-06
 Expose bool in expressions

### LINKED ISSUE #369: Allow `builtins.bool` in xacro expressions

SuperJappie08:
Fixes #369 

(Supersedes #370)


SuperJappie08:
@rhaschke, FYI, if this is ported to ROS2 and 'released' then it might still be able to be in the next ROS Jazzy sync.
(It was mentioned to be at the end of this week in the last PMC meeting.)

Issue #374, opened 2025-08-27
ros2 version can crash in `process_file` if `input_file_name` is not of `string` type.


acalhoon:
The `PickNikRobotics/launch_param_builder` module has passed a `pathlib.Path` object to `xacro.process_file` for a long time now. Recent changes on the `xacro` module ros2 branch attempt to sanitize the `input_file_name` for rendering inside of an XML comment assume that the input will be a `string` which results in an exception.

I created [this issue](https://github.com/PickNikRobotics/launch_param_builder/issues/16) in their repository, but it's not clear to me whether `xacro.process_file` should perform the explicit conversion to string within the `re.sub` call.

Curious to hear the maintainers' opinions.

Happy to create a pull request if necessary.

rhaschke:
Merged via https://github.com/ros/xacro/commit/3d061ea4a96af48fb96cd340980584d4fb6eee66 and https://github.com/ros/xacro/commit/46cb0515713d39bc5c46dde2a5e7b432ba533693.

PR #375, opened 2025-08-28
Fix path regex substitution

Noel215:
Hi,

[This commit](https://github.com/ros/xacro/commit/17ecece8961c21e82255e294704486e39de1145a) broke the method `process_file` when the `input_file_name` argument if of type `pathlib.Path` instead of `os.path` (string).

This PR fixes that issue, adding a test for compatibility with both path types

Already reported in the following issues:
- https://github.com/ros/xacro/issues/374
- https://github.com/PickNikRobotics/launch_param_builder/issues/16


acalhoon:
Thanks for the quick patch and for creating the PR for me!

rhaschke:
Thanks for the PR. Merged via 3d061ea4a96af48fb96cd340980584d4fb6eee66 and 46cb0515713d39bc5c46dde2a5e7b432ba533693.

